PR #18, opened 2020-10-03
Condition for status code used to throw errors changed

### LINKED ISSUE #15: (Easy) Change a successful refresh call to be when the status code is < 300 as opposed to === 200
In the [onUnauthorisedResponse function](https://github.com/supertokens/supertokens-website/blob/master/lib/ts/index.ts#L467), in index.ts, we check if the response from the refresh API call is === 200. This results in an issue where if the user instead sends something like 201, it stops working.

We want to change that to check if the response is < 300.

sudiptog81:
Tries to address and close #15.

Is this the only change required in `onUnauthorisedResponse`?

### DIFF
--- a/lib/ts/index.ts
+++ b/lib/ts/index.ts
@@ -514,7 +514,7 @@ export async function onUnauthorisedResponse(
                         setIDToCookie("remove", websiteRootDomain);
                     }
                 }
-                if (response.status > 300) {
+                if (response.status >= 300) {
                     throw response;
                 }
                 if (getIDFromCookie() === undefined) {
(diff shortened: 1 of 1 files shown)


rishabhpoddar:
Hi @sudiptog81 . Yes. But it should be `>= 300`

sudiptog81:
> 
> 
> Hi @sudiptog81 . Yes. But it should be `>= 300`

Fixed it in the latest commit.

rishabhpoddar:
Thanks! I have added you to the contributor's list on supertokens-core 👍

PR #27, opened 2020-12-10
use getWindowOrThrow

NkxxkN:
- Replace all `window` access with `getWindowOrThrow()`
- Replace `document` with `getWindowOrThrow().document`


lib/ts/utils.ts
```diff
@@ -210,3 +210,13 @@ export function validateAndNormaliseInputOrThrowError(options: InputType): Norma
         autoAddCredentials
     };
 }
+
+export function getWindowOrThrow(): any {
+    if (typeof window === "undefined") {
```
rishabhpoddar, line 215: Any reason to use `typeof window === "undefined"` as opposed to `window === undefined`?

NkxxkN: Yes, if we do `window === undefined`, trying to access window object throws the `Reference Error` which we are trying to avoid.

PR #65, opened 2021-06-28
Listen on IPv4 and IPv6 adapters

seniorquico:
Listening on `0.0.0.0` explicitly binds to IPv4. However, listening on `::` binds to both IPv4 and IPv6. On my network, `localhost.org` resolved to `::1` by default.

Note: I don't have an IPv4-only network on which to test this change. I cannot verify that it doesn't break IPv4-only usage scenarios.

### DIFF
--- a/test/server/index.js
+++ b/test/server/index.js
@@ -266,4 +266,4 @@ app.use(async (err, req, res, next) => {
 });
 
 let server = http.createServer(app);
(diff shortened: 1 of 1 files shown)


rishabhpoddar:
Thanks for this PR @seniorquico . I have also added you as a contributor in the supertokens-core readme :)

Issue #76, opened 2021-09-01
Forward error from refresh API to the client as well


rishabhpoddar:
For axios interceptor, if the refresh API throws a 500 error, then the use of axios does not get a proper error result back

PR #85, opened 2021-10-01
Axios refresh error

porcellus:
## Summary of change
We are now calling the session refresh endpoint through axios so that we can reject with an axios response object if the refresh call gets an unexpected error.

## Related issues
- #76 

## Test Plan
Added an extra test case

## Documentation changes
There are some properties that could be set in preAPIHook that axios does not support. I don't think these be set (or be useful) on this request (e.g., cache & referrerPolicy)

   - In `package.json`
   - In `package-lock.json`
   - In `lib/ts/version.ts`

## Remaining TODOs:
- [ ] new version of supertokens-website docs


test/server/index.html
```diff
@@ -1,5 +1,6 @@
 <html>
 <script src="https://unpkg.com/axios/dist/axios.min.js"></script>
+<script src="https://unpkg.com/axios-mock-adapter/dist/axios-mock-adapter.js"></script>
```
rishabhpoddar, line 3: This will also require a change in the backend SDK's integration test's index.html files.

lib/ts/axios.ts
```diff
@@ -275,6 +275,25 @@ export default class AuthHttpRequest {
             return await httpCall(config);
         }
 
+        // We make refresh calls through axios so that we have axios response object in case it makes it out of the API.
+        // This happens if there is an unexpected error during refresh (not sessionExpiredStatusCode).
+        const axiosFetch = async (url: string, config?: RequestInit) => {
+            const res = await axios({
+                url,
+                validateStatus: null,
```
rishabhpoddar, line 283: Write a comment about this flag here.

CHANGELOG.md
```diff
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [8.2.4] - 2021-10-01
```
rishabhpoddar, line 7: Make this 9.0.0, and also say why this is a breaking change (Cause the error API is different)

rishabhpoddar (changes requested):

test/server/index.html
porcellus: supertokens/supertokens-node#187

rishabhpoddar:
@porcellus also need change to golang repo (like what you did in nodejs repo)

rishabhpoddar (approved):

PR #87, opened 2021-10-01
Moved axios to dev-dependency and using fetch to call refresh

porcellus:
## Summary of change
Moved axios to dev-dependency and using fetch to call refresh

## Related issues
- #86 

## Test Plan
I updated the relevant test

## Documentation changes
N/A

   - In `package.json`
   - In `package-lock.json`
   - In `lib/ts/version.ts`


test/axios2.test.js
```diff
@@ -144,7 +192,9 @@ describe("Axios AuthHttpRequest class tests", function() {
                     exception = ex;
                 }
                 assertNotEqual(exception, undefined);
-                assertNotEqual(exception.response, undefined);
+                assertEqual(exception.status, 500);
+                assertNotEqual(exception.data, undefined);
+                assertEqual(exception.data.message, "test");
```
rishabhpoddar, line 197: The `exception` format still seems to be incorrect. We need to make it `exception.response.status`, `exception.response.data` in case of response being `500`. See https://stackoverflow.com/a/51768316/3867175

lib/ts/axios.ts
```diff
@@ -277,21 +277,18 @@ export default class AuthHttpRequest {
...
+            return {
+                data: await response.json(),
+                status: response.status,
+                statusText: response.statusText,
+                headers: response.headers,
+                config: {
+                    url: response.url,
+                    headers: response.headers
+                },
+                request: undefined
```
rishabhpoddar, line 290: I think that this is the interface for a valid response from an `axios` call, and not an error from an `axios` call. We need to use the interface of an `error`?

rishabhpoddar (changes requested):

rishabhpoddar:
Lists of tests we need to make for the new supertokens-website change:
- [ ] calling an API from fetch and that results in an error, causes behaviour of as if our interceptor doesn’t exist.
- [ ] calling an API from axios and that results in an error, causes behaviour of as if our interceptor doesn’t exist.
- [ ] in fetch, errors (including 401 errors) returned from refresh API call, results in a behaviour of as if the user had called the refresh API themselves and are handling the error.
- [ ] in axios, (including 401 errors) errors returned from refresh API call, results in a behaviour of as if the user had called the refresh API themselves and are handling the error.

When I say “behaviour”, I mean the output of the axios / fetch call from the user’s point of view - weather to expect it to return or throw, and the shape of the object returned / thrown. This must include those cases in which fetch throws error instead of returning it.

We should then also test the SDK in supertokens-auth-react. Both, manual testing, and SDK tests. For manual testing, we should:
- [ ] Start the frontend, without the backend server running, and do a fetch & axios call to an endpoint and see if error returned / thrown is as expected.
- [ ] Start the frontend without any cookies, start the backend, and see that the initial call to refresh fails with a 401, and IRTFrontend is set to “remove” (as expected)
- [ ] Start the frontend, not logged in, start the backend, make a call to an API that requires session verification and see that the output of the API call is a 401 (either returned or thrown based on using fetch or axios). This should then result in an automatic refresh call which should also result in a 401 which is then propogated to the user (either returned or thrown based on if fetch or axios is used)
- [ ] Start the frontend, not logged in, start the backend. Call an API from the frontend that returns a 500 error and see that that error is properly propogated to the user.
- [ ] Start the frontend, logged in, start the backend. Call an API from the frontend that returns a 500 error and see that that error is properly propogated to the user.


What we are editing right now is pretty core to the session recipe. So we must be very careful to make sure there are no bugs.

rishabhpoddar:
@porcellus let's get on a call to discuss these changes please.

rishabhpoddar (changes requested):
- Remove try / catch around fetch
- Do not need to throw responseObj in fetch
- Axios calling refresh unnecessarily if a session does not exist and the API returns a 401
- Add comments for your changes, especially the tricky parts.
- For `fetch` tests, add one where you make a call to an API without any session and check that the error is from that API call and not the refresh API call (i.e., it doesn't attempt to refresh at all and returns the error as is).
- FIx the test for axios where the refresh endpoint returns a 401.

test/axios2.test.js
```diff
@@ -145,6 +274,10 @@ describe("Axios AuthHttpRequest class tests", function() {
                 }
                 assertNotEqual(exception, undefined);
                 assertNotEqual(exception.response, undefined);
+                // assertEqual(exception.config.url, `${BASE_URL}/auth/session/refresh`);
```
rishabhpoddar, line 277: Please uncomment?

test/axios2.test.js
```diff
@@ -145,6 +274,10 @@ describe("Axios AuthHttpRequest class tests", function() {
...
+                assertEqual(exception.response.status, 401);
+                assertNotEqual(exception.response.data, undefined);
+                assertEqual(exception.response.data.message, "try refresh token");
```
rishabhpoddar, line 280: refresh API never returns this message.

rishabhpoddar (changes requested):

rishabhpoddar:
@porcellus the first two items in https://github.com/supertokens/supertokens-website/pull/87#issuecomment-932484551 are not checked. Any reason?

porcellus:
Hmm, I thought I checked them.

rishabhpoddar (approved):

PR #90, opened 2021-10-28
Adding bundle size checking as a PR action

porcellus:
## Summary of change
Adding bundle size checking as a PR action

## Related issues
- supertokens/supertokens-core#178

## Test Plan
Build change only

## Documentation changes
N/A

   - In `package.json`
   - In `package-lock.json`
   - In `lib/ts/version.ts`


rishabhpoddar:
I prefer if there is check for this in the CICD pipeline as well - when we add a dev tag. And if it crosses a certain limit, then it should fail the CICD.

Issue #116, opened 2022-04-09
Use with RTK Query - unable to assign request headers (specifically `Content-Type` and `rid`)


anmilleriii:
Edit: this does pertain to ST's implementation of a `fetch` wrapper.

~~Possibly out of scope of ST since this is for use with a third-party library (Redux Toolkit (RTK) Query), please close if so.~~

As described in this RTK issue: https://github.com/reduxjs/redux-toolkit/issues/2205#issuecomment-1094141663:

The expected behavior is the headers should be included in the request if included in any of the ways described (using `prepareHeaders`, assigning object literal to `headers`, or assigning the header within the `header` key in an endpoint's `query`).

The observed behavior headers are not able to be assigned (including `rid` and `Content-Type` to HTTP requests made with `fetchBaseQuery` (see https://redux-toolkit.js.org/rtk-query/api/fetchBaseQuery#using-fetchbasequery) w/SuperTokens. This results in the request body not being parsed/sent correctly.

Thought I would open here since RTK Query is gaining traction, thanks!

phryneas:
This is not specific to RTK Query - it seems that `supertokens` creates a wrapper around `fetch` that does not comply with the standard.

The first argument to ]`fetch`](https://fetch.spec.whatwg.org/#fetch-method) has to be a [RequestInfo](https://fetch.spec.whatwg.org/#requestinfo) which is either a `string`, or a `Request` class instance. 

The supertokens wrapper only accepts a string, RTK Query calls it with a `Request` object.

rishabhpoddar:
Oh I see! We shall fix this right away. Thanks for the issue.

rishabhpoddar:
On seeing the code again, our `fetch` wrapper's first param is indeed a `RequestInfo`: https://github.com/supertokens/supertokens-website/blob/master/lib/ts/recipeImplementation.ts#L9.

But I will still see why this is not working

phryneas:
It seems to come from you calling 
```js
                        return originalFetch(url, {
                            ...config
                        });
```

Reading the spec, that will lead to a call to `new Request(url, {...config})` internally.

Assuming that `url` in that case is a `Request` object that already has headers, going by the "new Request(input, init) constructor steps" described in the standard, any `headers` in `config` would overwrite the existing `headers` in `url` completely instead of merging them.

rishabhpoddar:
[...]

PR #117, opened 2022-04-10
fix: apply headers set in the first param of fetch

porcellus:
## Summary of change
Fixes how we construct headers in intercepted fetch requests.

## Related issues
- #116

## Test Plan

Added a test that checks if headers are applied through request interception.

## Documentation changes

N/A

   - In `package.json`
   - In `package-lock.json`
   - In `lib/ts/version.ts`


lib/ts/fetch.ts
```diff
@@ -200,22 +200,17 @@ export default class AuthHttpRequest {
                 // we read this here so that if there is a session expiry error, then we can compare this value (that caused the error) with the value after the request is sent.
                 // to avoid race conditions
                 const preRequestIdToken = await getIdRefreshToken(true);
-                let configWithAntiCsrf: RequestInit | undefined = config;
+                const clonedHeaders = new Headers(
+                    config !== undefined && config.headers !== undefined ? config.headers : url.headers
+                );
```
rishabhpoddar, line 205: The `url` object might have other params set too that might need to be copied over to the `config` object? Cause if someone is using that, then they are likely not using the `config` object.

So we will have to copy over all the properties to the `config` object

test/fetch.test.js
```diff
@@ -1859,6 +1859,42 @@ describe("Fetch AuthHttpRequest class tests", function() {
...
+            assert.equal(req2.headers()["rid"], "anti-csrf");
+            assert.equal(req2.headers()["asdf2"], "123");
+
+            assert.equal(req3.headers()["rid"], "anti-csrf");
+            assert.equal(req3.headers()["asdf"], undefined);
+        } finally {
+            await browser.close();
+        }
+    });
+
```
rishabhpoddar, line 1897: If you set up a demo app that uses the redux/toolkit lib and uses the `fetchBaseQuery` (like in [here](https://redux-toolkit.js.org/rtk-query/api/fetchBaseQuery#using-fetchbasequery)), and then try and trigger the refresh session flow, you will get an error on the API retry (post refreshing) saying `Cannot read properties of undefined (reading 'subscribe')`. Try and debug / fx that too please.

rishabhpoddar (changes requested):

lib/ts/fetch.ts
porcellus: 1. As far as I can see from the spec, there are no other properties that'd be overwritten.
2. I don't see why we need this? If we do that, it might be better to copy and modify the request object instead of the init config.

rishabhpoddar: > As far as I can see from the spec, there are no other properties that'd be overwritten.

If this is true, then we don't need to copy over anything else.

rishabhpoddar (approved):

rishabhpoddar (approved):

PR #121, opened 2022-05-07
feat: add access token updated event

porcellus:
## Summary of change
Add an event that signals if the access token has been updated to support updating the access token payload in the session context of the react lib.

## Related issues
- supertokens/supertokens-auth-react/issues/397

## Test Plan

Added a test case for both Axios and fetch.

## Documentation changes

- [ ] TODO: update `handle-event.mdx` in the session recipe

   - In `package.json`
   - In `package-lock.json`
   - In `lib/ts/version.ts`


lib/ts/axios.ts
```diff
@@ -179,6 +181,9 @@ export function responseInterceptor(axiosInstance: any) {
                 let frontToken = response.headers["front-token"];
                 if (frontToken !== undefined) {
                     await FrontToken.setItem(frontToken);
+                    if (!isNewSession) {
+                        onTokenUpdate();
+                    }
```
rishabhpoddar, line 186: This makes sense, however there could be another (maybe better) way of doing this: In `FrontToken.setItem`, we check if a token already exists (so a session exists), and then in case the new access token payload is different than what is saved, we fire this event. We would have to be careful about not doing that during session refresh.

What do you think?

test/axios2.test.js
```diff
@@ -1271,4 +1271,102 @@ describe("Axios AuthHttpRequest class tests", function() {
...
+                });
+            });
+            assert.deepEqual(logs, [
+                "SESSION_CREATED",
+                "LOGIN_FINISH",
+                "ACCESS_TOKEN_PAYLOAD_UPDATED",
+                "UPDATE1_FINISH",
+                "REFRESH_SESSION",
+                "ACCESS_TOKEN_PAYLOAD_UPDATED",
+                "REFRESH_FINISH",
```
rishabhpoddar, line 1360: Shouldn't this not have the `ACCESS_TOKEN_PAYLOAD_UPDATED` event?

rishabhpoddar (changes requested):

rishabhpoddar:
@porcellus there are a few conflicts too.

test/axios2.test.js
porcellus: I later renamed this event to `ACCESS_TOKEN_UPDATED` since that is more accurate, and the next call after a refresh does update the access token.

CHANGELOG.md
```diff
@@ -9,6 +9,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+
+### Fixes
+
+-   Setting headers in first param of fetch (https://github.com/supertokens/supertokens-website/issues/116)
+
+## [10.0.9] - 2022-03-18
+### Adds
+-   New FDI support (v1.13)
+-   Workflow to verify if pr title follows conventional commits
+
```
rishabhpoddar, line 25: Why is the changelog showing from previous PRs too?

test/axios2.test.js
rishabhpoddar: > and the next call after a refresh does update the access token.

I don't understand this. You removed the `ACCESS_TOKEN_PAYLOAD_UPDATED`, therefore the older `ACCESS_TOKEN_PAYLOAD_UPDATED` was incorrect?

CHANGELOG.md
rishabhpoddar: nvm.. ignore this comment

lib/ts/axios.ts
```diff
@@ -361,6 +361,8 @@ export default class AuthHttpRequest {
                         }
                         let frontToken = response.headers["front-token"];
                         if (frontToken !== undefined) {
+                            // this is also done by the response interceptor so
+                            // not firing the access token updated event
```
rishabhpoddar, line 365: This comment is not necessary anymore.

lib/ts/fetch.ts
```diff
@@ -558,7 +565,9 @@ export async function setIdRefreshToken(idRefreshToken: string | "remove", statu
         AuthHttpRequest.config.onHandleEvent({
             action: "SESSION_CREATED"
         });
+        return true;
     }
+    return false;
```
rishabhpoddar, line 570: We do not need to return a bool from this function anymore.

lib/ts/fetch.ts
```diff
@@ -685,5 +698,13 @@ export async function setFrontToken(frontToken: string | undefined) {
...
     }
 
+    const oldToken = await getFrontToken();
+    if (oldToken !== null && frontToken !== undefined) {
+        const oldPayload = parseFrontToken(oldToken).up;
+        const newPayload = parseFrontToken(frontToken).up;
+        if (JSON.stringify(oldPayload) !== JSON.stringify(newPayload)) {
+            onTokenUpdate();
+        }
+    }
```
rishabhpoddar, line 708: This may get triggered during session refresh as well - if the refresh API causes a change in the access token payload. I think it should fire even during session refresh - while from auth-react's point of view this will cause a double update, semantically, firing the `ACCESS_TOKEN_PAYLOAD_UPDATED ` event still makes sense.

What do you think?

rishabhpoddar: We should also add a test for this where a session refresh causes a change in the access token payload and then on session refresh, we check that `ACCESS_TOKEN_PAYLOAD_UPDATED` is fired.

rishabhpoddar (changes requested):

lib/ts/fetch.ts
porcellus: Yep, I think that makes sense. It already works that way, so I've added a test for it.

porcellus:
>I don't understand this. You removed the ACCESS_TOKEN_PAYLOAD_UPDATED, therefore the older ACCESS_TOKEN_PAYLOAD_UPDATED was incorrect?

So: I've renamed the event to `ACCESS_TOKEN_UPDATED` which is what I was checking in the test, and that was correct - and later I've removed it, since it doesn't make sense if we are only firing the event if the payload was updated (and I've reverted the rename)

test/axios2.test.js
```diff
@@ -1271,4 +1271,182 @@ describe("Axios AuthHttpRequest class tests", function() {
...
+                "ACCESS_TOKEN_PAYLOAD_UPDATED",
+                "UPDATE3_FINISH",
+                "SIGN_OUT"
+            ]);
+        } finally {
+            await browser.close();
+        }
+    });
+
+    it("test when ACCESS_TOKEN_PAYLOAD_UPDATED is fired", async function() {
```
rishabhpoddar, line 1372:
```suggestion
    it("test ACCESS_TOKEN_PAYLOAD_UPDATED when updated with handle", async function() {
```

rishabhpoddar (changes requested):

PR #136, opened 2022-06-14
feat(sessionclaims): adding validateClaims, base classes and types

porcellus:
## Summary of change

Adding base classes, types for claims, and `validateClaims`.

## Related issues
- https://github.com/supertokens/supertokens-node/pull/278
- https://github.com/supertokens/supertokens-auth-react/pull/400

## Test Plan
Tests will be done in a separate PR, but:
- [ ] Tests for `validateClaims`
- [ ] Tests for `PrimitiveClaim`
- [ ] Tests for `BooleanClaim`

## Documentation changes
N/A for now this is for internal use

   - In `package.json`
   - In `package-lock.json`
   - In `lib/ts/version.ts`


rishabhpoddar (changes requested):

lib/ts/index.ts
```diff
@@ -100,4 +137,9 @@ export let attemptRefreshingSession = AuthHttpRequest.attemptRefreshingSession;
 export let doesSessionExist = AuthHttpRequest.doesSessionExist;
 export let addAxiosInterceptors = AuthHttpRequest.addAxiosInterceptors;
 export let signOut = AuthHttpRequest.signOut;
+export const validateClaims = AuthHttpRequest.validateClaims;
 export { RecipeInterface, InputType };
+
+export { ClaimValidationError, ClaimValidationResult, SessionClaimValidator } from "./types";
+export { PrimitiveClaim } from "./claims/primitiveClaim";
+export { BooleanClaim } from "./claims/booleanClaim";
```
rishabhpoddar, line 145: should export them as "supertokens-website/claims"

rishabhpoddar (changes requested):

lib/ts/index.ts
```diff
@@ -91,6 +91,43 @@ export default class AuthHttpRequest {
             userContext: getNormalisedUserContext(input === undefined ? undefined : input.userContext)
         });
     };
+
```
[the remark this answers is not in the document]
porcellus: Done in #138

lib/ts/index.ts
rishabhpoddar: On second thought, this is not needed because when importing supertokens-website via a script tag, im not sure how users would use these claims. 

On this subject, even if these are here, how will users use this when importing via script tag?

rishabhpoddar (changes requested):

lib/ts/types.ts
```diff
@@ -110,3 +110,32 @@ export type RecipeInterface = {
 
     signOut: (input: { userContext: any }) => Promise<void>;
 };
+
```
rishabhpoddar, line 113: should the recipe interface for sessions also have a `getGlobalClaimValidators` function which the user can override? Just like how it's done on the backend?

rishabhpoddar (changes requested):

lib/ts/index.ts
porcellus: `supertokens.BooleanClaim`

lib/ts/types.ts
```diff
@@ -109,6 +109,11 @@ export type RecipeInterface = {
     doesSessionExist: (input: { userContext: any }) => Promise<boolean>;
 
     signOut: (input: { userContext: any }) => Promise<void>;
+
+    validateClaims: (input: {
+        claimValidators: SessionClaimValidator[];
+        userContext?: any;
+    }) => Promise<ClaimValidationError[] | undefined>;
```
rishabhpoddar, line 116: Having the return type be undefined is a bit weird? When the same thing can also be signified with an empty array being returned? Cause now, the user will have to do:
```
let errors = await validateClaims(...)

if (errors !==undefined && errors.length > 0) {
  ... handle error
}
```

As opposed to just doing `errors.length > 0` check

lib/ts/types.ts
rishabhpoddar: @porcellus missed this?

lib/ts/index.ts
```diff
@@ -91,6 +92,22 @@ export default class AuthHttpRequest {
...
+    static getInvalidClaimsFromResponse = async function(
+        response: AxiosResponse | Response
```
rishabhpoddar, line 97: missing optional userContext in input

lib/ts/types.ts
```diff
@@ -109,4 +116,46 @@ export type RecipeInterface = {
     doesSessionExist: (input: { userContext: any }) => Promise<boolean>;
 
     signOut: (input: { userContext: any }) => Promise<void>;
+
+    getInvalidClaimsFromResponse(input: { response: AxiosResponse | Response }): Promise<ClaimValidationError[]>;
```
rishabhpoddar, line 120: Missing userContext - I know that it won't be needed as such, but we have to be consistent with other functions.

lib/ts/types.ts
rishabhpoddar: @porcellus missed this?

lib/ts/recipeImplementation.ts
```diff
@@ -137,6 +145,62 @@ export default function RecipeImplementation(recipeImplInput: {
...
+            if ("body" in input.response) {
+                body = await input.response.clone().json();
+            } else {
+                body = input.response.data;
+            }
+
+            return body.claimValidationErrors;
+        },
+
+        getGlobalClaimValidators: async function(input: {
```
rishabhpoddar, line 163: this function is not actually being used anywhere? Shuoldn't we use it when validateClaim is called?

lib/ts/recipeImplementation.ts
```diff
@@ -137,6 +145,62 @@ export default function RecipeImplementation(recipeImplInput: {
...
+            claimValidatorsAddedByOtherRecipes: SessionClaimValidator[];
+        }) {
+            return input.claimValidatorsAddedByOtherRecipes;
+        },
+
+        validateClaims: async function(input: {
+            claimValidators: SessionClaimValidator[];
```
rishabhpoddar, line 170: shouldn't this be similar to how it's in node? Instead of taking an array here, we should take an optional function like `overrideGlobalClaimValidators`?

rishabhpoddar (changes requested):

lib/ts/recipeImplementation.ts
rishabhpoddar: We should also take a lock while validating (as discussed)

lib/ts/index.ts
```diff
@@ -91,6 +92,52 @@ export default class AuthHttpRequest {
...
+            userContext: normalisedUserContext
+        });
+        const claimValidators =
+            overrideGlobalClaimValidators !== undefined
+                ? overrideGlobalClaimValidators(globalClaimValidators, normalisedUserContext)
+                : globalClaimValidators;
+
+        if (claimValidators.length === 0) {
+            return [];
+        }
```
rishabhpoddar, line 126: Shouldn't this go inside the recipe implementation file? Similar to how it's done in the node SDK. Or maybe we should make the Node SDK such that it deosn't happen in the recipe implementation.

I think I prefer the approach taken in this SDK cause then it keeps the interface of `validateClaims` cleaner, for when the user is overriding the function.

lib/ts/index.ts
```diff
@@ -91,6 +92,52 @@ export default class AuthHttpRequest {
...
+    }): Promise<ClaimValidationError[]> {
+        return AuthHttpRequestFetch.recipeImpl.getInvalidClaimsFromResponse({
+            response: input.response,
+            userContext: getNormalisedUserContext(input.userContext)
+        });
+    };
+
+    static validateClaims = (
+        overrideGlobalClaimValidators?: (
+            globalClaimValidators: SessionClaimValidator[],
```
rishabhpoddar, line 108:
```suggestion
            globalClaimValidators: SessionClaimValidator[] | Promise<SessionClaimValidator[]>,
```

lib/ts/index.ts
```diff
@@ -91,6 +92,52 @@ export default class AuthHttpRequest {
...
+            userContext: any
+        ) => SessionClaimValidator[],
+        userContext?: any
+    ): Promise<ClaimValidationError[]> | ClaimValidationError[] => {
+        const normalisedUserContext = getNormalisedUserContext(userContext);
+        const claimValidatorsAddedByOtherRecipes = AuthHttpRequest.getClaimValidatorsAddedByOtherRecipes();
+        const globalClaimValidators = AuthHttpRequestFetch.recipeImpl.getGlobalClaimValidators({
```
rishabhpoddar, line 115:
```suggestion
        const globalClaimValidators = await AuthHttpRequestFetch.recipeImpl.getGlobalClaimValidators({
```

lib/ts/types.ts
```diff
@@ -109,4 +116,48 @@ export type RecipeInterface = {
...
+
+    validateClaims: (input: {
+        claimValidators: SessionClaimValidator[];
+        userContext: any;
+    }) => Promise<ClaimValidationError[]>;
+
+    getGlobalClaimValidators(input: {
+        claimValidatorsAddedByOtherRecipes: SessionClaimValidator[];
+        userContext: any;
+    }): SessionClaimValidator[];
```
rishabhpoddar, line 133:
```suggestion
    }): SessionClaimValidator[] | Promise<SessionClaimValidator[]>;
```

rishabhpoddar: Just like in the node SDK

lib/ts/index.ts
```diff
@@ -91,6 +92,52 @@ export default class AuthHttpRequest {
...
+            claimValidatorsAddedByOtherRecipes,
+            userContext: normalisedUserContext
+        });
+        const claimValidators =
+            overrideGlobalClaimValidators !== undefined
+                ? overrideGlobalClaimValidators(globalClaimValidators, normalisedUserContext)
```
rishabhpoddar, line 121:
```suggestion
                ? await overrideGlobalClaimValidators(globalClaimValidators, normalisedUserContext)
```

lib/ts/index.ts
```diff
@@ -91,6 +92,52 @@ export default class AuthHttpRequest {
...
+        });
+    };
+
+    static addClaimValidatorFromOtherRecipe = (builder: SessionClaimValidator) => {
+        AuthHttpRequest.claimValidatorsAddedByOtherRecipes.push(builder);
+    };
+
+    static getClaimValidatorsAddedByOtherRecipes = (): SessionClaimValidator[] => {
+        return AuthHttpRequest.claimValidatorsAddedByOtherRecipes;
+    };
```
rishabhpoddar, line 140: there are not being exported as named exports below?

lib/ts/claims/primitiveClaim.ts
```diff
@@ -0,0 +1,75 @@
...
+    constructor(config: PrimitiveClaimConfig) {
+        this.id = config.id;
+        this.refresh = config.refresh;
+    }
+
+    getValueFromPayload(payload: any, _userContext?: any): ValueType {
+        return payload[this.id] === undefined ? payload[this.id].v : undefined;
+    }
+
+    getLastFetchedTime(payload: any, _userContext?: any): Date | undefined {
```
rishabhpoddar, line 21: should return `number | undefined`

rishabhpoddar (changes requested):

lib/ts/index.ts
rishabhpoddar: We decided to keep this synchronous cause it makes it easier for us to implement SSR

lib/ts/index.ts
rishabhpoddar, line 140: You can export these through supertokens-website/utils, and on web-js side, via supertokens-web-js/recipe/session/utils (if required at all)

rishabhpoddar (changes requested):

lib/ts/index.ts
porcellus: I also prefer this method.
In the node SDK, I moved it into a separate function on the `recipeImplementation` as we've discussed.

PR #141, opened 2022-07-08
feat(sessionclaims): support for email verification claim

porcellus:
## Summary of change

Add email verification claim

## Related issues

-  https://github.com/supertokens/supertokens-node/pull/349
-  https://github.com/supertokens/supertokens-web-js/pull/48

## Test Plan

Done separately

## Documentation changes

Done separately

   - In `package.json`
   - In `package-lock.json`
   - In `lib/ts/version.ts`


lib/ts/recipeImplementation.ts
```diff
@@ -199,7 +199,8 @@ export default function RecipeImplementation(recipeImplInput: {
                 if (!validationRes.isValid) {
                     errors.push({
                         validatorId: validator.id,
-                        reason: validationRes.reason
+                        reason: validationRes.reason,
+                        redirectPath: validationRes.redirectPath
```
rishabhpoddar, line 203: Im not sure I quite understand this - where does `redirectPath` come from? If it comes from the backend, how would the backend know what the path is?

rishabhpoddar (changes requested):

lib/ts/recipeImplementation.ts
rishabhpoddar: Please figure out a way to remove redirectToPath from website and web-js and adding that specific field in auth-react. Some ideas we discussed:
- Adding a callback to the auth-react's `validateGlobalClaimsAndRedirect` which will be called when claim validations fail.
- Making the error type generic here.

lib/ts/recipeImplementation.ts
```diff
@@ -199,7 +199,8 @@ export default function RecipeImplementation(recipeImplInput: {
...
                     });
                 }
```
rishabhpoddar, line 205: We should add a test to enforce that the claims sent to `validateClaims` array are indeed checked in the same order as they are in the array. Which means that the error[] returned should have the errors in the same order as well.

rishabhpoddar (changes requested):

lib/ts/recipeImplementation.ts
porcellus: done in #143

porcellus:
Decided to go in a different direction that doesn't need an interface change here.

PR #143, opened 2022-07-15
test(sessionclaims): add tests for session claims

porcellus:
## Summary of change

Added tests for session claims

## Related issues
- 

## Test Plan
AuthHttpRequest claim handling
    AuthHttpRequest validateClaims
      SessionClaimValidatorStore interactions
        ✓ should call override with empty list if nothing is added
        ✓ should call override with the appropriate args when claims are added
        ✓ should return empty array if all global validators pass
        ✓ should return non-empty array if a global validator fails
      validateClaims
        ✓ should return empty array if no validators are present
        ✓ should return empty array if all validators pass
        ✓ should return empty array if all validators after the override
        ✓ should return non-empty array if a local validator fails
        ✓ should return failure reasons in the order returned from the override
        ✓ should call refresh only if shouldRefresh returns true
        ✓ should refresh all claims before validation

  Session claims error handling
    fetch
      ✓ should return a parseable body
    axios
      ✓ should throw with a parseable body

## Documentation changes

   - In `package.json`
   - In `package-lock.json`
   - In `lib/ts/version.ts`

### DIFF
--- a/test/server/index.js
+++ b/test/server/index.js
@@ -275,6 +275,23 @@ app.post(
     }
 );
 
+app.post(
+    "/session-claims-error",
+    (req, res, next) =>
+        verifySession({
+            overrideGlobalClaimValidators: () => [
+                {
+                    id: "test-claim-failing",
+                    shouldRefetch: () => false,
+                    validate: () => ({ isValid: false, reason: { message: "testReason" } })
+                }
+            ]
+        })(req, res, next),
+    async (req, res) => {
+        res.json({});
+    }
+);
+
 app.use("/testing", async (req, res) => {
     let tH = req.headers["testing"];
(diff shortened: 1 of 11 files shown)


test/server/index.js
```diff
@@ -275,6 +275,23 @@ app.post(
     }
 );
 
+app.post(
```
rishabhpoddar, line 278: Perhaps you need to edit `featureFlags` in the server since this server is run with older versions of the node SDK as well. But this is only needed if the new version of supertokens-website is still compatible with older FDIs

rishabhpoddar (changes requested):

rishabhpoddar (approved):

PR #145, opened 2022-07-22
feat(sessionclaims): add primitive array claim

porcellus:
## Summary of change
Adds a claim that handles arrays of JSON primitives

## Related issues
- 

## Test Plan
Same code is tested in the node PR (but I'll add extra tests separately)

## Documentation changes
Done separately

   - In `package.json`
   - In `package-lock.json`
   - In `lib/ts/version.ts`


lib/ts/claims/primitiveArrayClaim.ts
```diff
@@ -0,0 +1,228 @@
...
+                        };
+                    }
+                    for (const [i, v] of Array.from(val.entries())) {
+                        if (claimVal[i] !== v) {
+                            return {
+                                isValid: false,
+                                reason: { message: "wrong value", expectedValue: val, actualValue: claimVal }
+                            };
+                        }
+                    }
```
rishabhpoddar, line 220: Does strictEquals imply that the order also has to be the same? It's ambiguous, therefore maybe we can get rid of this function, and ask users to use something like `includesAll(target) && getValueFromPayload().length === target.length`.

rishabhpoddar (changes requested):

lib/ts/claims/primitiveArrayClaim.ts
porcellus: It does, so doing `includesAll` + length check would be different. Anyway, there is no use-case for it yet, so I'll just remove it.

rishabhpoddar (approved):

PR #149, opened 2022-08-11
feat: session claims

porcellus:
## Summary of change

Adds support for session claims

## Related issues
- supertokens/supertokens-auth-react/pull/400
- supertokens/supertokens-web-js/pull/57

## Test Plan
Added tests

## Documentation changes
Done separately

   - In `package.json`
   - In `package-lock.json`
   - In `lib/ts/version.ts`


utils/sessionClaimValidatorStore/index.js
```diff
@@ -0,0 +1,20 @@
...
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License. 
+*/
+"use strict";
+function __export(m) {
+    for (var p in m) if (!exports.hasOwnProperty(p)) exports[p] = m[p];
+}
+exports.__esModule = true;
+__export(require("../../lib/build/utils/sessionClaimValidatorStore"));
```
rishabhpoddar, line 20:
```suggestion
exports.__esModule = true;
let d = require("../../lib/build/utils/sessionClaimValidatorStore");

if (d.default !== undefined) {
    __export(d);
} else {
    __export({
        default: d,
        ...d,
    });
}
```

rishabhpoddar: The sessionClaimValidatorStore does not have a default export, but the .d.ts file associated with this JS file implies that it has.. so we change this to add a default export.

lib/ts/fetch.ts
```diff
@@ -298,6 +298,11 @@ export default class AuthHttpRequest {
                     }
                     logDebugMessage("doRequest: Retrying original request");
                 } else {
+                    if (response.status === AuthHttpRequest.config.invalidClaimStatusCode) {
```
rishabhpoddar, line 301: Are you sure that you don't want to check for this in the refreshing session part as well? What if the user overrides that on the backend to do claim validation and return this type of error?

lib/ts/claims/primitiveClaim.ts
```diff
@@ -0,0 +1,75 @@
...
+                    return isValid
+                        ? { isValid: isValid }
+                        : {
+                              isValid,
+                              reason: { message: "wrong value", expectedValue: val, actualValue: claimVal }
+                          };
+                }
+            };
+        },
+        hasFreshValue: (val: ValueType, maxAgeInSeconds: number, id?: string): SessionClaimValidator => {
```
rishabhpoddar, line 43: We do not want to have hasFreshValue anymore - just like on the backend.

lib/ts/claims/primitiveClaim.ts
```diff
@@ -0,0 +1,75 @@
...
+    getValueFromPayload(payload: any, _userContext?: any): ValueType {
+        return payload[this.id] !== undefined ? payload[this.id].v : undefined;
+    }
+
+    getLastFetchedTime(payload: any, _userContext?: any): number | undefined {
+        return payload[this.id] !== undefined ? payload[this.id].t : undefined;
+    }
+
+    validators = {
+        hasValue: (val: ValueType, id?: string): SessionClaimValidator => {
```
rishabhpoddar, line 26: should also take in a maxAgeSeconds param like how the backend does.

lib/ts/claims/primitiveClaim.ts
```diff
@@ -0,0 +1,75 @@
...
+export type PrimitiveClaimConfig = {
+    id: string;
+    refresh: (userContext?: any) => Promise<void>;
+};
+
+export class PrimitiveClaim<ValueType> {
+    public readonly id: string;
+    public readonly refresh: SessionClaimValidator["refresh"];
+
+    constructor(config: PrimitiveClaimConfig) {
```
rishabhpoddar, line 12: Should take a maxAgeSeconds thing as well just like how the backend does.

lib/ts/claims/primitiveArrayClaim.ts
```diff
@@ -0,0 +1,186 @@
+import { SessionClaimValidator } from "../types";
+
+export type PrimitiveArrayClaimConfig = {
+    id: string;
+    refresh: (userContext?: any) => Promise<void>;
+};
+
```
rishabhpoddar, line 7: please double check this file to match the backend as well.. in case it deosn't

rishabhpoddar (changes requested):

lib/ts/fetch.ts
rishabhpoddar: oh right
[22:04](https://supertokens.slack.com/archives/D02B887A2MQ/p1661358843883719)
I forgot this this morning
[22:04](https://supertokens.slack.com/archives/D02B887A2MQ/p1661358890232979)
umm yeah, so I'm sure. a claim validation error coming from the refresh endpoint doesn't really make sense
[22:05](https://supertokens.slack.com/archives/D02B887A2MQ/p1661358920280709)
since you'd have no way of refreshing the access token payload
[22:06](https://supertokens.slack.com/archives/D02B887A2MQ/p1661358964338409)
and since there is no way of updating the token but it fails validation the only sensible way to handle this is to remove the session... so it's essentially a 401.

rishabhpoddar (approved):

PR #153, opened 2022-08-24
Xmlhttprequest interception

rishabhpoddar:
## Summary of change
Adds XMLHttpRequest inrterception so that axios and HttpClient can "just work"

## Related issues
- Link to issue1 here
- Link to issue1 here

## Test Plan

## Documentation changes

   - In `package.json`
   - In `package-lock.json`
   - In `lib/ts/version.ts`

- [ ] Testing with HttpClient
- [ ] Check if we can just assign to `actual.onerror`, actual.onload`, `actual.onreadystatechange` and `actual.onloadend` without checking about if those features are available.
- [ ] Check default value of `withCredentials`
- [ ] Migration method - if this method is buggy, how can users disable this entirely and fallback on older axios interception?


lib/ts/xmlhttprequest.ts
```diff
@@ -0,0 +1,429 @@
...
+
+        // we do not provide onerror cause that is fired only on
+        // network level failures and nothing else. If a status code is > 400,
+        // then onload and onreadystatechange are called.
+
+        self.onload = null;
+        self.onreadystatechange = null;
+        self.onloadend = null;
+
+        actual.onload = function(this: XMLHttpRequestType, ev: ProgressEvent<EventTarget>) {
```
rishabhpoddar, line 57: not sure if or when this is called - do we need to even modify this?

lib/ts/xmlhttprequest.ts
```diff
@@ -0,0 +1,429 @@
...
+                })
+            ) {
+                requestHeaders = requestHeaders.filter(i => {
+                    return i.name !== name;
+                });
+            }
+            requestHeaders.push({ name, value });
+            actual.setRequestHeader(name, value);
+        };
+
```
rishabhpoddar, line 347: How to handle progress events? Cause the actual instance will say progress is finished when it's done, but then if we retry, then progress will start again.

lib/ts/xmlhttprequest.ts
```diff
@@ -0,0 +1,429 @@
...
+        let customGetterValues: { [key: string]: any } = {};
+
+        // we do not provide onerror cause that is fired only on
+        // network level failures and nothing else. If a status code is > 400,
+        // then onload and onreadystatechange are called.
+
+        self.onload = null;
+        self.onreadystatechange = null;
+        self.onloadend = null;
+
```
rishabhpoddar, line 56: how can we handle cases where a user is downloading a file which is guarded with verifySession?

PR #159, opened 2022-09-16
feat: header based auth

porcellus:
## Summary of change

Implement header based auth

## Related issues
- 

## Test Plan
- Made sure old test-cases still ran without change for backwards compatibility
- Copied much of the session tests and modified them to use headers

## Documentation changes

Not yet.

   - In `package.json`
   - In `package-lock.json`
   - In `lib/ts/version.ts`


porcellus:
Some TODOs will be done in separate PR (removing `id-refresh-token`)

lib/ts/fetch.ts
```diff
@@ -279,15 +279,20 @@ export default class AuthHttpRequest {
                     logDebugMessage("doRequest: rid header was already there in request");
                 }
 
+                await setTokenHeaders(clonedHeaders);
```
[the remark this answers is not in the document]
rishabhpoddar: EDIT: We actually do want to do the above cause if we don't then migration would be a problem (from cookie to header based). So what we will do is to add these credentials and anti-csrf header all the time, but then after a month or so of release, we will set the autoAddCredentials to false by default in case mode is headers

lib/ts/fetch.ts
```diff
@@ -421,8 +429,11 @@ export async function onUnauthorisedResponse(
                 );
 
                 logDebugMessage("onUnauthorisedResponse: Refresh call ended");
+
+                await saveTokensFromHeaders(response);
+
```
rishabhpoddar, line 434: We should think about old -> new migration path here:
- Remove old tokens like sIRTFrontend
- If method is header based, and anti-csrf token exists, we want to remove that as well.

lib/ts/fetch.ts
```diff
@@ -279,15 +279,20 @@ export default class AuthHttpRequest {
...
+                await setTokenHeaders(clonedHeaders);
+
                 logDebugMessage("doRequest: Making user's http call");
                 let response = await httpCall(configWithAntiCsrf);
                 logDebugMessage("doRequest: User's http call ended");
 
-                const idRefreshToken = response.headers.get("id-refresh-token");
+                await saveTokensFromHeaders(response);
+
+                const idRefreshToken = response.headers.get("st-id-refresh-token");
```
rishabhpoddar, line 290: id refresh will be removed

lib/ts/fetch.ts
```diff
@@ -636,55 +626,130 @@ export async function getIdRefreshToken(tryRefresh: boolean): Promise<IdRefreshT
...
+
+        const accessToken = response.headers.get("st-access-token");
+        if (accessToken) {
+            const [value, expiry] = accessToken.split(";");
+            await setToken("access", value, Number.parseInt(expiry));
+        }
+    }
+}
+
+export async function setIdRefreshToken(idRefreshToken: string | "remove", statusCode: number) {
```
rishabhpoddar, line 745: this function should go away

lib/ts/fetch.ts
```diff
@@ -345,6 +350,8 @@ export default class AuthHttpRequest {
 }
 
 const ID_REFRESH_TOKEN_NAME = "sIRTFrontend";
```
rishabhpoddar, line 352: this constant goes

lib/ts/fetch.ts
```diff
@@ -386,23 +393,24 @@ export async function onUnauthorisedResponse(
                     // means that some other process has already called this API and succeeded. so we need to call it again
                     return { result: "RETRY" };
                 }
```
rishabhpoddar, line 395: There are places where we check the value of id refresh token to know if a session exists (instead of calling the doesSessionExist), so that needs to be changed to use the front token and the lastRefreshAttempt time stamp

lib/ts/fetch.ts
```diff
@@ -573,29 +584,8 @@ type IdRefreshTokenType =
 // else we return undefined.
 export async function getIdRefreshToken(tryRefresh: boolean): Promise<IdRefreshTokenType> {
     logDebugMessage("getIdRefreshToken: called");
```
rishabhpoddar, line 586: tryRefresh is false:
   - if front token exists -> EXISTS
   - if front token does not exist, lastRefreshTime exists -> NOT_EXISTS
   - Else -> MAY_EXISTS

try refresh is true:
   - if front token exists -> EXISTS
   - if front token does not exist, lastRefreshTime exists -> NOT_EXISTS
   - Else -> We try refreshing the session

rishabhpoddar (changes requested):

lib/ts/fetch.ts
```diff
@@ -636,55 +621,133 @@ export async function getIdRefreshToken(tryRefresh: boolean): Promise<IdRefreshT
...
-                }`
-            );
+    }
+    return undefined;
+}
+
+async function setTokenHeadersIfRequired(clonedHeaders: Headers, addRefreshToken: boolean = false) {
+    if (AuthHttpRequest.config.tokenTransferMethod === "header") {
+        logDebugMessage("setTokenHeadersIfRequired: adding ';header' to the rid header");
+        clonedHeaders.set("rid", clonedHeaders.get("rid") + ";header");
```
rishabhpoddar, line 701: - I think we had decided to extract this into another header, instead of appending to rid
- The default on the backend is header. So this condition should be about cookies

rishabhpoddar (changes requested):

porcellus:
Will be done in #173

PR #173, opened 2022-11-20
feat(session rework): remove id refresh token

porcellus:
## Summary of change
Remove refresh-id-token

## Related issues
- https://supertokens.com/docs/contribute/decisions/session/0011

## Test Plan
Added separately

## Documentation changes
Done separately

   - In `package.json`
   - In `package-lock.json`
   - In `lib/ts/version.ts`


lib/ts/fetch.ts
```diff
@@ -567,167 +566,203 @@ export async function onInvalidClaimResponse(response: ResponseWithBody) {
...
+    // cookie. This way, when we query for this token, we will return
+    // undefined (see getLocalSessionState), and not refresh the session
+    // unnecessarily.
+    logDebugMessage(`setToken: saved ${tokenType} token into cookies`);
+    return storeInCookies(name, value, expiry);
+}
+
+function storeInCookies(name: string, value: string, expiry: number) {
+    let expires = "Fri, 31 Dec 9999 23:59:59 GMT";
+    if (value !== "remove" && expiry !== Number.MAX_SAFE_INTEGER) {
```
rishabhpoddar, line 642: Maybe this function should not care about if the value is `remove`? Cause when you set front token in the FrontToken class, you call the setFrontToken function with `undefined` which just calls this with `value = ""`

lib/ts/fetch.ts
```diff
@@ -567,167 +566,203 @@ export async function onInvalidClaimResponse(response: ResponseWithBody) {
...
+        // we must always respect this expiry and not set it to infinite
+        // cause this ties into the session's lifetime. If we set this
+        // to infinite, then a session may not exist, and this will exist,
+        // then for example, if we check a session exists, and this says yes,
+        // then if we getAccessTokenPayload, that will attempt a session refresh which will fail.
+        // Another reason to respect this is that if we don't, then signOut will
+        // call the API which will return 200 (no 401 cause the API thinks no session exists),
+        // in which case, we will not end up firing the SIGN_OUT on handle event.
```
rishabhpoddar, line 650: > // If we set this
        // to infinite, then a session may not exist, and this will exist,
        // then for example, if we check a session exists, and this says yes
Doesn't deosSessionExist also check the expiry of front-token in the token itself? If this is set to infinite, then if the access token expires, the front-token's expiry will force a refresh.. and then doesSessionExist will get the correct value - so is this comment correct?

lib/ts/fetch.ts
```diff
@@ -567,167 +566,203 @@ export async function onInvalidClaimResponse(response: ResponseWithBody) {
...
+        if (last !== undefined) {
+            return last.split(";").shift();
         }
     }
+    return undefined;
+}
 
-    const { status } = await getIdRefreshToken(false);
+async function setTokenHeadersIfRequired(clonedHeaders: Headers, addRefreshToken: boolean = false) {
+    const transferMethod = AuthHttpRequest.config.tokenTransferMethod;
```
rishabhpoddar, line 693: in this function, we do not check if `transferMethod` is = `header` before setting the authorization header - that is intentional right? If so, please comment this here linking to the relevant ADR(s)

lib/ts/fetch.ts
```diff
@@ -567,167 +566,203 @@ export async function onInvalidClaimResponse(response: ResponseWithBody) {
...
+        const [value, expiry] = refreshToken.split(";");
+        await setToken("refresh", value, Number.parseInt(expiry));
+    }
+
+    const accessToken = response.headers.get("st-access-token");
+    if (accessToken) {
+        logDebugMessage("saveTokensFromHeaders: saving new access token");
+        const [value, expiry] = accessToken.split(";");
+        await setToken("access", value, Number.parseInt(expiry));
+    }
```
rishabhpoddar, line 731: Thinking about this, it would be easier if the backend just didn't set the expiry of these tokens and the frontend just set them to 100 years. That would then make the communication and implementation of non frontend SDK folks easier where they just have to set the value of `st-refresh-token` or `st-access-token` without the splitting stuff

lib/ts/fetch.ts
```diff
@@ -281,19 +294,17 @@ export default class AuthHttpRequest {
...
                 logDebugMessage("doRequest: Making user's http call");
                 let response = await httpCall(configWithAntiCsrf);
                 logDebugMessage("doRequest: User's http call ended");
 
-                const idRefreshToken = response.headers.get("id-refresh-token");
-                if (idRefreshToken) {
-                    logDebugMessage("doRequest: Setting sIRTFrontend: " + idRefreshToken);
-                    await setIdRefreshToken(idRefreshToken, response.status);
-                }
+                await saveTokensFromHeaders(response);
```
rishabhpoddar, line 303: Earlier we used to only store the front-token and anti-csrf if it was a non 401 response, but now we do this regardless - does this have any side effect?

lib/ts/fetch.ts
```diff
@@ -567,167 +566,203 @@ export async function onInvalidClaimResponse(response: ResponseWithBody) {
...
+        let last = parts.pop();
+        if (last !== undefined) {
+            return last.split(";").shift();
         }
     }
+    return undefined;
+}
 
-    const { status } = await getIdRefreshToken(false);
+async function setTokenHeadersIfRequired(clonedHeaders: Headers, addRefreshToken: boolean = false) {
```
rishabhpoddar, line 692:
```suggestion
async function setAuthorizationHeaderIfRequired(clonedHeaders: Headers, addRefreshToken: boolean = false) {
```

lib/ts/fetch.ts
```diff
@@ -567,167 +566,203 @@ export async function onInvalidClaimResponse(response: ResponseWithBody) {
...
+    const transferMethod = AuthHttpRequest.config.tokenTransferMethod;
 
-    logDebugMessage("setIdRefreshToken: setting: " + idRefreshToken);
-    await setIDToCookie(idRefreshToken, AuthHttpRequest.config.sessionScope);
+    logDebugMessage("setTokenHeadersIfRequired: Adding st-auth-mode header: " + transferMethod);
+    clonedHeaders.set("st-auth-mode", transferMethod);
```
rishabhpoddar, line 696: it's a little strange that this header is being set here but `rid` and `anti-csrf` headers are being set outside this function. Maybe move setting this header outside this function as well?

lib/ts/fetch.ts
```diff
@@ -428,22 +433,24 @@ export async function onUnauthorisedResponse(
...
+                // There is a case where the FE thinks the session is valid, but backend doesn't get the tokens.
+                // In this event, session expired error will be thrown and the frontend should remove this token
+                if (isUnauthorised && response.headers.get("front-token") === null) {
+                    FrontToken.setItem("remove");
                 }
+
+                fireSessionUpdateEventsIfNecessary(
+                    preRequestLSS.status === "EXISTS",
+                    response.status,
+                    isUnauthorised ? "remove" : response.headers.get("front-token")
```
rishabhpoddar, line 452:
```suggestion
                    isUnauthorised && response.headers.get("front-token") === null ? "remove" : response.headers.get("front-token")
```

lib/ts/fetch.ts
```diff
@@ -428,22 +433,24 @@ export async function onUnauthorisedResponse(
...
-                    }
+
+                await saveTokensFromHeaders(response);
+
+                logDebugMessage("onUnauthorisedResponse: Refresh status code is: " + response.status);
+
+                const isUnauthorised = response.status === AuthHttpRequest.config.sessionExpiredStatusCode;
+
+                // There is a case where the FE thinks the session is valid, but backend doesn't get the tokens.
+                // In this event, session expired error will be thrown and the frontend should remove this token
```
rishabhpoddar, line 444: is this still possible? Can we eliminate this state somehow? If not, it's OK - but just asking

lib/ts/fetch.ts
```diff
@@ -428,22 +433,24 @@ export async function onUnauthorisedResponse(
...
-                    removeIdRefreshToken = false;
-                }
-                if (response.status === AuthHttpRequest.config.sessionExpiredStatusCode) {
-                    logDebugMessage("onUnauthorisedResponse: Refresh status code is: " + response.status);
-                    // there is a case where frontend still has id refresh token, but backend doesn't get it. In this event, session expired error will be thrown and the frontend should remove this token
-                    if (removeIdRefreshToken) {
-                        await setIdRefreshToken("remove", response.status);
-                    }
+
+                await saveTokensFromHeaders(response);
```
rishabhpoddar, line 437: earlier, we used to save front-token and ant-csrf token only if status code was < 300. But now we do it even if status code is >=300. Are there any side effects?

rishabhpoddar (changes requested):

lib/ts/fetch.ts
```diff
@@ -107,7 +107,7 @@ export class FrontToken {
         logDebugMessage("FrontToken.getTokenInfo: called");
```
rishabhpoddar, line 107: also need a function (exposed via the index file) to get the access token in case of header based auth. This would be especially useful for users who are not using a library which has our interceptors, or are using stuff like websockets or trpc

rishabhpoddar (changes requested):

lib/ts/recipeImplementation.ts
```diff
@@ -44,6 +44,22 @@ export default function RecipeImplementation(recipeImplInput: {
...
+                console.warn(
+                    "Since XMLHttpRequest is added automatically and supports axios by default, you can just remove addAxiosInterceptors from your code."
+                );
+                console.warn(
+                    "If you want to continue using our axios interceptor, you can override addXMLHttpRequestInterceptor with an empty function."
+                );
+
+                logDebugMessage("addAxiosInterceptors: not adding, because XHR interceptors are already in place");
+                return;
+            }
```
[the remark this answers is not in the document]
porcellus: I mean this is just a warning, and the Axios interceptor is already circumvented anyway if the XHR interceptor was added.
Still, I think this will have to be a breaking change in the website repo anyway (and BE is breaking for sure) but I'll check (and add an explicit FE test about continuing an old session)

lib/ts/fetch.ts
```diff
@@ -567,167 +566,203 @@ export async function onInvalidClaimResponse(response: ResponseWithBody) {
...
-            status: "NOT_EXISTS"
-        };
-    }
-
-    if (token === undefined) {
-        logDebugMessage("getIdRefreshToken: is undefined");
-        let response: IdRefreshTokenType = {
+export async function getLocalSessionState(tryRefresh: boolean): Promise<LocalSessionState> {
+    logDebugMessage("getLocalSessionState: called");
+
```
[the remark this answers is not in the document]
porcellus: Not on the FE - we still call the API, which returns a 401 in case we need to call migration, which is handled the same as a normal refresh.

lib/ts/fetch.ts
```diff
@@ -889,5 +832,57 @@ export async function setFrontToken(frontToken: string | undefined) {
...
+    }
+}
+
+export function fireSessionUpdateEventsIfNecessary(
+    wasLoggedIn: boolean,
+    status: number,
+    frontTokenHeader: string | null | undefined
+) {
+    if (frontTokenHeader === undefined || frontTokenHeader === null) {
+        // The access token (and the session) hasn't been updated.
```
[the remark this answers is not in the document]
porcellus: It's not obvious from this function, but the 401 branch is handled separately by the callers. Maybe I could restructure this, but I wanted to keep it the same where possible to minimize this PR.
That being said, I think there are code repetition issues that made making this change harder than necessary that could later warrant a restructuring.

lib/ts/fetch.ts
porcellus: It is intentional: the frontend preference might be to use header or cookie-based auth, but that will only take effect after a new session is created, and by default, we can use the existing one until the user logs out.
I'll add this as a comment.

lib/ts/fetch.ts
porcellus: I'd prefer if the refresh token still had the expiration (although it's not critical), and we should keep them consistent.
Also, for cookie-based sessions, we'd still set this expiration on the BE so I think it makes sense to also have them in headers.

lib/ts/fetch.ts
porcellus: This should be really rare:
1. someone clearing HTTP only cookies
2. some issue before/during saving the front-token
Still, we can't eliminate it completely

lib/ts/fetch.ts
```diff
@@ -889,5 +832,57 @@ export async function setFrontToken(frontToken: string | undefined) {
...
+        await storeInCookies(FRONT_TOKEN_NAME, "", 0);
+    } else {
+        await storeInCookies(FRONT_TOKEN_NAME, frontToken, Number.MAX_SAFE_INTEGER);
+    }
+}
+
+export function fireSessionUpdateEventsIfNecessary(
+    wasLoggedIn: boolean,
+    status: number,
+    frontTokenHeader: string | null | undefined
```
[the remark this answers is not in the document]
porcellus: I don't mind updating this, but the front-token header is only present in the response

lib/ts/recipeImplementation.ts
```diff
@@ -113,7 +113,24 @@ export default function RecipeImplementation(recipeImplInput: {
         },
         doesSessionExist: async function(_: { userContext: any }): Promise<boolean> {
             logDebugMessage("doesSessionExist: called");
-            return (await getLocalSessionState(true)).status === "EXISTS";
+
+            const tokenInfo = await FrontToken.getTokenInfo();
+
+            // The above includes getIdRefreshToken(true), which would call refresh if the FE cookies were cleared for some reason
```
rishabhpoddar, line 119: comment is old

lib/ts/recipeImplementation.ts
```diff
@@ -113,7 +113,24 @@ export default function RecipeImplementation(recipeImplInput: {
...
+            if (tokenInfo === undefined) {
+                logDebugMessage("doesSessionExist: access token does not exist locally");
+                return false;
+            }
+
+            if (tokenInfo.ate < Date.now()) {
+                logDebugMessage("doesSessionExist: access token expired. Refreshing session");
+
+                const preRequestIdToken = await getLocalSessionState(false);
+                const refresh = await onUnauthorisedResponse(preRequestIdToken);
```
rishabhpoddar, line 129: variable name should not have IdToken

lib/ts/index.ts
```diff
@@ -143,6 +143,15 @@ export default class AuthHttpRequest {
             userContext: getNormalisedUserContext(userContext)
         });
     };
+
+    static getAccessToken = async (input?: { userContext?: any }): Promise<string | undefined> => {
+        // This takes care of refreshing the access token if necessary.
+        await AuthHttpRequestFetch.recipeImpl.getAccessTokenPayloadSecurely({
+            userContext: getNormalisedUserContext(input === undefined ? undefined : input.userContext)
+        });
```
rishabhpoddar, line 151: if no session exists, this will throw an error - shouldn't we return `undefined` in this case from this function instead?

There are three cases:
1) session does not exist -> throw error?
2) session exists, cookie being used, so access token can't be read by frontend JS -> return undefined?
3) session exists, access token exists -> return token.

So what im saying is that case (1) and (2) should return `undefined`. Do you think that's a good idea? Or maybe case (1) should return `undefined`, but case (2) should throw an error? This way, if cookie transfer method is being used, user should not call this function.

lib/ts/index.ts
```diff
@@ -143,6 +143,15 @@ export default class AuthHttpRequest {
             userContext: getNormalisedUserContext(userContext)
         });
     };
+
+    static getAccessToken = async (input?: { userContext?: any }): Promise<string | undefined> => {
```
rishabhpoddar, line 147: this funtion shuold be in recipe interface and recipe implementation should have this code.

lib/ts/index.ts
rishabhpoddar, line 151: im not sure if implicitly doing this is good. If tomorrow the impl of `getAccessTokenPayloadSecurely` changes ever so slightly, then we may not rememeber to also check this function. Instead, you should do what `getAccessTokenPayloadSecurely` does.

lib/ts/fetch.ts
```diff
@@ -134,7 +134,11 @@ export class FrontToken {
 
     static async setItem(frontToken: string) {
         // We update the refresh attempt info here as well, since this means that we've updated the session in some way
-        // maybe through a custom endpoint...
+        // This could be both by a refresh call or if the access token was updated in a custom endpoint
+        // By saving every time the access token has been updated, we cause an early retry if
+        // another request has failed with a 401 with the previous access token and the token still exists.
```
rishabhpoddar, line 139: - Why do we need to cause an early retry? If refreshing is taking place, and during that, the access token was updated, it could still mean that the access token is expired cause updating the access token payload doesn't change it's expiry.
- If we must do this, then we should probably change the `saveRefreshAttempt` function to something like `saveAccessTokenUpdateTime` -> this would store the last time the access token was updated or refreshed (it;'s a better name given the context).

lib/ts/fetch.ts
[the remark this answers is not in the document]
rishabhpoddar: Ah okay. Then maybe just write this scenario and your reply as a comment in the code - for anyone who sees this function later on and thinks about what i thought about

lib/ts/fetch.ts
```diff
@@ -689,27 +699,29 @@ async function getFromCookies(name: string) {
...
 
-    logDebugMessage("setTokenHeaders: adding existing tokens as header");
+    const accessToken = await getTokenForHeaderAuth("access");
+    const refreshToken = await getTokenForHeaderAuth("refresh");
 
-    const token = await getTokenForHeaderAuth(addRefreshToken ? "refresh" : "access");
-    if (token) {
+    // We don't add the refresh token because that's only required by the refresh call which is done with fetch
+    // Still, we only add the Authorization header if both are present, because we are planning to add an option to expose the
+    // access token to the frontend while using cookie based auth
```
rishabhpoddar, line 714:
```suggestion
    // access token to the frontend while using cookie based auth - so that users can get the access token to use 
    // in a (for example) websocket connection whilst using cookie based transfer method for http protocol.
    // In this case, the refresh token will be in cookies and not accessible to the frontend, and the access token
    // will also be in the cookies. So we don't want to auto add the access token as part of the authorization header
    // if the refresh token is not accessible to the frontend.
```

lib/ts/fetch.ts
rishabhpoddar: hmm. Debatable, but ok. let's keep it the way it is cause it's implemented this way already.

lib/ts/fetch.ts
```diff
@@ -382,34 +397,38 @@ export async function onUnauthorisedResponse(
...
                         "onUnauthorisedResponse: Retrying early because pre and post id refresh tokens don't match"
                     );
                     // means that some other process has already called this API and succeeded. so we need to call it again
                     return { result: "RETRY" };
                 }
-                let headers: any = {};
-                if (preRequestIdToken.status === "EXISTS") {
-                    const antiCsrfToken = await AntiCsrfToken.getToken(preRequestIdToken.token);
+
+                await saveRefreshAttempt();
```
rishabhpoddar, line 412: Im not sure if we should call this here. I think we should call this after refresh is successful cause:
- User is using cookie based auth in safari.
- Safari clears all frontend cookies - including last refresh time (refresh token in httpOnly cookies still exists)
- This causes a refresh attempt to happen which saves the last refresh time to now.
- The refresh API is called but it fails (500 error for example) resulting in none of the cookies of the front token to be set.
- The if `getLocalSessionState` is called again, it will return `NOT_EXISTS` even if the refresh token exists in cookies.

lib/ts/fetch.ts
rishabhpoddar: @porcellus missed this?

rishabhpoddar (changes requested):

lib/ts/index.ts
porcellus: Getting the token from storage is already customizable through `windowHandler`, and I think it'd be confusing:
- We are not using this function internally when getting the token
- We'd have to add an overrideable `setAccessToken` (and maybe others for refresh token, etc.)

lib/ts/index.ts
porcellus: I think returning undefined in 1 and 2 should be fine.

lib/ts/fetch.ts
```diff
@@ -224,7 +242,7 @@ export default class AuthHttpRequest {
                 doNotDoInterception = !shouldDoInterceptionBasedOnUrl(
                     WindowHandlerReference.getReferenceOrThrow().windowHandler.location.getOrigin(),
                     AuthHttpRequest.config.apiDomain,
```
rishabhpoddar, line 244: check if the auth header added by the user is the same as the one we have - and if it is, then we remove the user added header and just let the interceptors take care of adding it

rishabhpoddar (changes requested):

lib/ts/index.ts
porcellus: I'll make this call `doesSessionExist` instead. I think if we change the implementation of that so that it no longer refreshes the session, then we'd have to check all places that call that function anyway. I'd prefer not to repeat the code for this reason.

lib/ts/fetch.ts
porcellus: - if request A gets a 401 and starts to acquire a lock, while request B updates the access token, it means that the access token is not expired since B could update it. If request B is executed on the server before A and the token expires between the two, then we make an extra retry call, but this should be rare enough.
- Without this, we'd miss user APIs creating (and, in rarer cases, refreshing) sessions if using cookies since we'd only see the front-token update.

The basic idea is that we knew we had an expired token, but that information is outdated since the token we'd be using for the retry isn't the same.

lib/ts/fetch.ts
porcellus: No, I just wanted to discuss this on the call, but we didn't end up doing it. (Also, this is basically the same issue as answered in another comment thread about saving tokens)
Basically: since the access token and front-token values are tied together, their updates should be close together as well. Otherwise, the front end could get into an inconsistent state.
This might not be an issue here, but it still seems better to always save these updates if the backend sends them. There could be some edge cases, e.g.: a custom endpoint calling refresh manually but returning a 401 may cause issues with anti-csrf, but I'm not a 100% sure about it.

lib/ts/fetch.ts
```diff
@@ -409,11 +424,9 @@ export async function onUnauthorisedResponse(
                     return { result: "RETRY" };
                 }
 
-                await saveRefreshAttempt();
-
```
rishabhpoddar: don't we need to update last access token update time after a successful refresh? If the refresh succeeds, then it will call the front-token update which will do this, but if the refresh fails (for example if the user is logged out), then it won't save this last access token update time and then it will keep calling refresh on each page load?

lib/ts/fetch.ts
```diff
@@ -254,6 +254,21 @@ export default class AuthHttpRequest {
             logDebugMessage("doRequest: Returning without interception");
             return await httpCall(config);
         }
+
+        const origHeaders = new Headers(
+            config !== undefined && config.headers !== undefined ? config.headers : url.headers
+        );
+
+        if (origHeaders.has("Authorization")) {
```
rishabhpoddar, line 262: what if the user has set lower case authorization?

lib/ts/fetch.ts
```diff
@@ -254,6 +254,21 @@ export default class AuthHttpRequest {
...
+            const accessToken = await getTokenForHeaderAuth("access");
+            if (accessToken !== undefined && origHeaders.get("Authorization") === `Bearer ${accessToken}`) {
+                logDebugMessage(
+                    "doRequest: Removing Authorization from user provided headers because it contains our access token"
+                );
+                origHeaders.delete("Authorization");
+            }
+        }
```
rishabhpoddar, line 270: should add a clear comment explaining why this is being done. Describe the scenario in detail.

lib/ts/fetch.ts
```diff
@@ -567,167 +595,201 @@ export async function onInvalidClaimResponse(response: ResponseWithBody) {
...
-            });
+    const accessToken = await getTokenForHeaderAuth("access");
+    const refreshToken = await getTokenForHeaderAuth("refresh");
+
+    // We don't add the refresh token because that's only required by the refresh call which is done with fetch
+    // Still, we only add the Authorization header if both are present, because we are planning to add an option to expose the
+    // access token to the frontend while using cookie based auth - so that users can get the access token to use
+    if (accessToken !== undefined && refreshToken !== undefined) {
+        // the Headers class normalizes header names so we don't have to worry about casing
+        if (clonedHeaders.has("Authorization")) {
```
rishabhpoddar, line 732: what if the user uses lowercase authorization?

lib/ts/fetch.ts
```diff
@@ -567,167 +595,201 @@ export async function onInvalidClaimResponse(response: ResponseWithBody) {
...
-            action: "SESSION_CREATED",
-            userContext: {}
-        });
+async function saveTokensFromHeaders(response: Response) {
+    logDebugMessage("saveTokensFromHeaders: Saving updated tokens from the response headers");
+
+    const refreshToken = response.headers.get("st-refresh-token");
+    if (refreshToken) {
+        logDebugMessage("saveTokensFromHeaders: saving new refresh token");
+        const [value, expiry] = refreshToken.split(";");
```
rishabhpoddar, line 749: On second thought, sending the expiry from the backend for header based auth (as part of the value for st-refersh-token or st-access-token) is very strange. Please just send the access token and refresh token without the `;` + time after.

The reason is cause it affects developer experience when using postman. People will call the login API and get back headets (since that the default auth mode), and then they will have to copy the access token and remove the `;<time>` part from it and then paste it in the authorization header. This is a bad DX.

On the frontend, we should just set the access and refresh token to be 100 years for header based auth.

rishabhpoddar (changes requested):

lib/ts/fetch.ts
```diff
@@ -304,19 +345,6 @@ export default class AuthHttpRequest {
...
-                        }
-                    }
-                    const frontToken = response.headers.get("front-token");
-                    if (frontToken) {
-                        logDebugMessage("doRequest: Setting sFrontToken: " + frontToken);
-                        await FrontToken.setItem(frontToken);
-                    }
                     return response;
                 }
             }
```
rishabhpoddar, line 350: do you also remove legacy sIRTFrontend token when updating to the newer session?

rishabhpoddar (changes requested):

lib/ts/fetch.ts
```diff
@@ -567,167 +595,201 @@ export async function onInvalidClaimResponse(response: ResponseWithBody) {
...
-                sessionExpiredOrRevoked: true,
-                userContext: {}
-            });
+    const accessToken = await getTokenForHeaderAuth("access");
+    const refreshToken = await getTokenForHeaderAuth("refresh");
+
+    // We don't add the refresh token because that's only required by the refresh call which is done with fetch
+    // Still, we only add the Authorization header if both are present, because we are planning to add an option to expose the
+    // access token to the frontend while using cookie based auth - so that users can get the access token to use
+    if (accessToken !== undefined && refreshToken !== undefined) {
```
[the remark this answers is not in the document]
porcellus: Yes, the token should still be present even if they expired: we are setting the access tokens with a 100-year expiry

lib/ts/fetch.ts
porcellus: the Headers class handles these as case insensitive.

lib/ts/fetch.ts
porcellus: We clear it during the access token update.

lib/ts/fetch.ts
porcellus: the Headers class handles these as case insensitive.

lib/ts/fetch.ts
porcellus: - if refresh returns anything that updates the front token (200, or 401 clearing tokens), FrontToken.setItem will save the update
- if refresh returns 401 that didn't update tokens (i.e.: FE thinks it has tokens, but BE didn't get them), then we clear FrontToken on line 473
- if we errored out for other reasons (i.e., we got a 500 response, and it didn't update any tokens), then I don't think calling the refresh again on the next page load is a problem

rishabhpoddar: One of the developer experience things is that the frontend UI should show even if the backend integration is not done. In this case, when the frontend loads up for the first time, it will get a 404 from the refresh endpoint, and each time the developer reloads the page or doesSessionExist is called, it will call the refresh API (until they have finished the backend integration). This might result in several calls to the refresh API in one go which can be weird.

If I am right, and if there is no harm in moving this to after the refresh has been called (even if it returned a 500 error), then please move it after.

rishabhpoddar: @porcellus ^^

porcellus: I think that's only appropriate in the 404 case
If:
- the FE cookies have been cleared (e.g.: in safari)
- the BE returns a 500 and we save the update cookie
Then on the next reload we will not call refresh and consider the user logged out even though they have valid tokens in the BE set cookies.

This is an edge-case, but can occur in production instead of a minor weird behaviour during setup.

rishabhpoddar: We need to know how many times refresh will get called if the backend is not started. We should also write a test for this case to make sure that refresh is not called 10 times or something in this case. Maybe 1-3 times is fine;

test/fetch.test.js
```diff
@@ -2181,12 +2163,11 @@ describe("Fetch AuthHttpRequest class tests", function() {
...
                         body: JSON.stringify({ message: "test" }),
                         headers: {
-                            "id-refresh-token": "remove",
                             "Set-Cookie": [
-                                "sIdRefreshToken=; Path=/; Expires=Thu, 01 Jan 1970 00:00:00 GMT; HttpOnly; SameSite=Lax",
-                                "sAccessToken=; Path=/; Expires=Thu, 01 Jan 1970 00:00:00 GMT; HttpOnly; SameSite=Lax",
+                                "sAccessToken=remove; Path=/; Expires=Fri, 31 Dec 9999 23:59:59 GMT; HttpOnly; SameSite=Lax",
                                 "sRefreshToken=; Path=/auth/session/refresh; Expires=Thu, 01 Jan 1970 00:00:00 GMT; HttpOnly; SameSite=Lax"
-                            ]
+                            ],
```
rishabhpoddar, line 2169: might wanna add a migration test

rishabhpoddar (changes requested):

lib/ts/fetch.ts
rishabhpoddar: (Keeping this comment open) - but the problem is resolved.

porcellus: in the test app we call 1-2 times depending on what page was opened

lib/ts/axios.ts
```diff
@@ -134,6 +134,9 @@ export async function interceptorFunctionRequestFulfilled(config: AxiosRequestCo
     const accessToken = getTokenForHeaderAuth("access");
     if (accessToken !== undefined) {
         if (configWithAntiCsrf.headers!.authorization === `Bearer ${accessToken}`) {
+            // We are ignoring the Authorization header set by the user in this case, because it would cause issues
+            // If we do not ignore this, then this header would be used even if the request is being retried after a refresh, even though it contains an outdated access token.
+            // This causes an infinite refresh loop.
             logDebugMessage(
                 "interceptorFunctionRequestFulfilled: Removing Authorization from user provided headers because it contains our access token"
             );
```
rishabhpoddar, line 142: I think calling `await setAuthorizationHeaderIfRequired(configWithAntiCsrf);` below is wrong?

rishabhpoddar (changes requested):

lib/ts/axios.ts
porcellus: we should definitely call it in `interceptorFunctionRequestFulfilled` otherwise the first request would always fail. After checking it out:
- The tests didn't cover this because `addXMLHttpRequestInterceptor` disabled the axios interceptor. I've changed this.
- This check should be done before going into the refresh loop in `doRequest`

Issue #175, opened 2022-11-23
doesSessionExists should check the front-token validity and refresh if necessary


porcellus:
Since we are not checking the front-token validity, a refresh call may be necessary even if `doesSessionExists` returns true. 
This means that future calls to things like `getAccessTokenPayloadSecurely` may throw if the refresh call throws.

leonard-henriquez:
Since I raised this issue and I know how to reproduce it, I would be delighted to test the fix to this issue when you have one !

rishabhpoddar:
hey @leonard-henriquez 

We have a fix for the problem, but it would be great if you can test it out in your setup and see if it solves the issue.

Can you run the following command on your project to get the new (unreleased) changes:
```bash
cd your_project
npm i git+https://github.com:supertokens/supertokens-website#fix/doesSessionExist_refresh
```

Rebuild your app and try again. To reproduce the issue, do the following:
- Set the access token lifetime to 15 seconds on the supertokens.com dashboard
- Login to your app and wait for 15 seconds
- Block the refresh endpoint
- Reload the page. 

If the issue is fixed, you should now see the login screen. 

Then unblock the refresh endpoint and reload the page again, and you should see the dashboard screen.

Thanks!

rishabhpoddar:
@leonard-henriquez did you get a chance to try the fix please? If not, that's OK, we will be looking to release this within the next 2 days anyway.

leonard-henriquez:
Sorry for the late response @rishabhpoddar , but I tired it and it works 🎉 !!

rishabhpoddar:
This has been fixed in version >= 14.0.3

PR #176, opened 2022-11-24
fix: make doesSessionExist refresh if front-token expired

porcellus:
## Summary of change
`doesSessionExist` should refresh the session if it detects an expired access token (through info in front-token)

## Related issues
- #175 

## Test Plan

Added a few new test cases. Other aspects of the function are tested during the interception tests.

## Documentation changes

N/A

   - In `package.json`
   - In `package-lock.json`
   - In `lib/ts/version.ts`


lib/ts/recipeImplementation.ts
```diff
@@ -97,7 +97,24 @@ export default function RecipeImplementation(recipeImplInput: {
...
+                logDebugMessage("doesSessionExist: access token does not exist locally");
+                return false;
+            }
+
+            if (tokenInfo.ate < Date.now()) {
+                logDebugMessage("doesSessionExist: access token expired. Refreshing session");
+
+                const preRequestIdToken = await getIdRefreshToken(false);
+                const refresh = await onUnauthorisedResponse(preRequestIdToken);
+                return refresh.result === "RETRY";
```
rishabhpoddar, line 114: See how refreshing is done in getAccessTokenPayloadSecurely in this file. Do it in a similar way, except that instead of `throw new Error("Could not refresh session");` we should return `false`?

rishabhpoddar (changes requested):

lib/ts/recipeImplementation.ts
porcellus: This is what I did here; I just replaced the `attemptRefreshingSession` since it throws if refreshing fails.

PR #178, opened 2022-12-09
feat!: add getWindowUnsafe to windowHandler

porcellus:
## Summary of change

Add getWindowUnsafe to windowHandler. This is used by a branch in supertokens-auth-react that removes the @emotion dependency

## Related issues
- supertokens-auth-react#624

## Test Plan

-

## Documentation changes

`windowHandler` is never mentioned in the docs.

   - In `package.json`
   - In `package-lock.json`
   - In `lib/ts/version.ts`


rishabhpoddar (approved):

lib/ts/utils/windowHandler/defaultImplementation.ts
```diff
@@ -134,6 +134,9 @@ export const defaultWindowHandlerImplementation: WindowHandlerInterface = {
     getDocument: function() {
         return getWindowOrThrow().document;
     },
+    getWindowUnsafe: function() {
+        return getWindowOrThrow().window;
+    },
```
rishabhpoddar, line 139: we need to check what happens is auth-react is used in hybrid apps like:
- capacitor
- electron

rishabhpoddar (changes requested):

lib/ts/utils/windowHandler/defaultImplementation.ts
porcellus: Checked both, and it worked without any issues.

rishabhpoddar (approved):

PR #186, opened 2023-01-13
feat: add lockFactory config

sublimator:
We are using SuperTokens with an extension that needs to refresh the tokens for a given domain that may be very infrequently browsed, or on the other hand may be browsed concurrently with the extension usage. 

We want to use supertokens-website from the background page of the extension, overriding the CookieHandler (to set tokens on the app domain using Web Extension apis) and customizing browser-tabs-tokens so that the extension can act as "just another tab" 

Wanted to get a feel if this PR is moving in the right direction and would have a chance to be merged. 

Thanks!


rishabhpoddar:
Thanks for this PR @sublimator .

@porcellus will review this in the coming days.

lib/ts/utils/windowHandler/index.ts
```diff
@@ -12,38 +12,45 @@
...
 export default class WindowHandlerReference {
-    private static instance?: WindowHandlerReference;
-
-    windowHandler: WindowHandlerInterface;
+    private static instance?: WindowHandlerReference
+
+    // Declare subset of WindowHandlerReference that is actually used,
+    // so implementors don't need to over-engineer.
+    windowHandler: {
+        location: Pick<
```
porcellus, line 24: We shouldn't limit this to a subset. This lib exports the window handler, and the other props are used by supertokens-web-js and supertokens-auth-react.

lib/ts/utils/windowHandler/index.ts
```diff
@@ -12,38 +12,45 @@
...
+            WindowHandlerInterface['location'],
+            'getHostName' | 'getOrigin'
+        >
+    }
 
     constructor(windowHandlerInput?: WindowHandlerInput) {
-        let windowHandlerFunc: WindowHandlerInput = original => original;
+        let windowHandlerFunc: WindowHandlerInput = original => original
```
porcellus, line 31: These updates do not match the coding style used elsewhere in the project; please run `npm run build-pretty`.
Also, I'm not sure if you've installed the commit hooks (check [here](https://github.com/supertokens/supertokens-website/blob/master/CONTRIBUTING.md#development-setup) if not)

lib/ts/types.ts
```diff
@@ -39,6 +40,13 @@ export type InputType = {
     enableDebugLogs?: boolean;
     apiDomain: string;
     apiBasePath?: string;
+    /**
+     * Rather than using browser-tabs-lock, allow for a Lock factory to be configured
+     * via SuperTokens.init(...). This can be used for example by a WebExtension that
+     * needs to update cookies for a domain that may or may not have an associated
+     * tab open.
```
porcellus, line 47:
```suggestion
     * This allows for a Lock factory to be configured, which defaults to browser-tabs-lock.
     * This can be used, for example, by a WebExtension that needs to update cookies for
     * a domain that may or may not have an associated tab open.
```

porcellus (changes requested):
This is a good change; there are just a few minor issues :)

lib/ts/utils/windowHandler/index.ts
sublimator: Fair, I shouldn't have committed this

lib/ts/utils/windowHandler/index.ts
sublimator: image

I think prettier might need to be updated ? Seems this is using 1.x while 2.x is out.

sublimator:
@porcellus 

I removed the limiting of the windowHandler type, ran prettier, updated it when it wasn't working with some of the syntax I used (`??` operator and `import type`), then tweaked the .prettierrc to be as close to 1.x as possible.   

If you are the maintainer (or one of), feel free to take over PR if it's more convenient for you.  

We just need some ability to customize the lock.

Thanks!

sublimator:
I made the lockFactory async. I'm not sure if we will need it, but it's more flexible, so why not?

porcellus (approved):
LGTM :)

package.json
```diff
@@ -20,7 +20,7 @@
     "mocha": "6.1.4",
```
rishabhpoddar, line 20: - Please update the changelog under the [unreleased] section
- Might want to expose the LockFactory in the `/utils/lockFactory` folder similar to how the [windowHandler is exposed](https://github.com/sublimator/supertokens-website/blob/nd-add-lock-factory-to-config-2023-01-13/utils/windowHandler/index.js)

rishabhpoddar (changes requested):

package.json
rishabhpoddar: The other changes look good. So once these two comments are addressed, i'll merge and release this.

sublimator:
@rishabhpoddar 

I'm not sure how/why I set the target branch to 15.0 and not master ?
Is that right? 

Anyway, I (hopefully) addressed your two comments.

rishabhpoddar (approved):

rishabhpoddar (approved):

PR #224, opened 2023-06-12
refactor: Do not throw if window is not defined during initialisation

nkshah2:
## Summary of change
- Changes the type of sessionTokenFrontendDomain to a function in the normalised config type
- Adds a check for XMLHttpRequest being undefined during init
- Refactors init to not throw an exception if window is not defined to allow for both server side and client side the initialise the SDK

## Related issues
- ...

## Documentation changes
WIP

   - In `package.json`
   - In `package-lock.json`
   - In `lib/ts/version.ts`

- [ ] Update CHANGELOG and package version


lib/ts/xmlhttprequest.ts
```diff
@@ -46,6 +46,11 @@ const XHR_EVENTS = [
 ] as const;
 
 export function addInterceptorsToXMLHttpRequest() {
+    // This is to prevent errors in server side environments where XMLHttpRequest is not defined
+    if (typeof XMLHttpRequest === "undefined") {
```
porcellus, line 50: Please add a similar check to all interceptors that return early (after reporting in debug logs that it'll skip adding the interceptor) in an SSR environment. Node 18 added a global `fetch`, which we are overriding if with this branch, with unintended side-effects.

PR #241, opened 2024-01-10
feat: Add DateProvider to fix clock skew issues

anku255:
## Summary of change

This PR introduced `DateProvider` that is used to fix clock skew issues between server and client. 

## Related issues

- https://github.com/supertokens/supertokens-web-js/pull/103

## Test Cases 
Tested the following cases locally simulated clock skew by changing the system clock

- [x] Standard network speed
- [x] Slow network speed
- [x] A large delay before updating the access token payload
- [x] A large delay after updating the access token payload

## Documentation changes

   - In `package.json`
   - In `package-lock.json`
   - In `lib/ts/version.ts`

- [x] Update version
- [x] Update changelog
- [x] Do the changes for axios and XMLHTTPRequest


lib/ts/fetch.ts
```diff
@@ -926,3 +931,43 @@ export function fireSessionUpdateEventsIfNecessary(
...
+        logDebugMessage(
+            "updateClientClockUsingFrontToken: FrontToken iat is undefined or not a number, skipping update"
+        );
+        return;
+    }
+
+    const estimatedServerTimeNow = tokenIssuedAt * 1000 + Math.floor(roundTripTime / 2);
+    const clientClockDeviationInMillis = estimatedServerTimeNow - Date.now();
+
+    AuthHttpRequest.recipeImpl.updateClientClockDeviation(clientClockDeviationInMillis);
```
porcellus, line 970: This is called `clock skew` in the literature, please update the naming scheme to match that. (incl. the comments)

lib/ts/recipeImplementation.ts
```diff
@@ -308,6 +309,16 @@ export default function RecipeImplementation(recipeImplInput: {
                     return domain === normalisedsessionDomain;
                 }
             }
+        },
+
+        updateClientClockDeviation: function (clientClockDeviationInMillis: number): void {
```
porcellus, line 314: What we want the users to customize is how we get the clock skew from the response. Ideally, this should get:
- the access token payload (which will include the `iat`)
- the response object if possible while being consistent across interceptors, or the headers at least.

It should return the detected clock skew as a number, which we will set in the `DateProvider`

lib/ts/types.ts
```diff
@@ -150,6 +150,8 @@ export type RecipeInterface = {
         apiDomain: string,
```
porcellus, line 150: The `DateProvider` needs to be customizable as well. Check out `windowHandler` in the input.

lib/ts/utils/dateProvider/defaultImplementation.ts
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+class DateProvider {
+    private clientClockDeviationInMillis: number = 0;
```
porcellus, line 17: We should try to save and re-load this from somewhere, otherwise this will have to be re-done on every refresh/full page navigation. Not 100% required, but please check if it's easily achievable

lib/ts/utils/dateProvider/index.ts
```diff
@@ -0,0 +1,57 @@
...
+    constructor(dateProviderInput?: DateProviderInput) {
+        let dateProviderFunc: DateProviderInput = original => original;
+        if (dateProviderInput !== undefined) {
+            dateProviderFunc = dateProviderInput;
+        }
+
+        this.dateProvider = dateProviderFunc(defaultDateProviderImplementation);
+    }
+
+    static init(dateProviderInput?: DateProviderInput): void {
```
porcellus, line 40: Where is this called? Something seems to be missing here.

test/sessionClaimValidatorRefresh.test.js
```diff
@@ -0,0 +1,129 @@
...
+        return cb();
+    } finally {
+        clock.restore();
+    }
+}
+
+describe("SessionClaimValidator Refresh", () => {
+    describe("Client and Server clock are in sync", () => {
+        before(function () {
+            DateProviderReference.init();
```
porcellus, line 36: Are the session claims tests passing?

test/sessionClaimValidatorRefresh.test.js
```diff
@@ -0,0 +1,129 @@
...
+    const clock = sinon.useFakeTimers({ now, shouldAdvanceTime: false });
+
+    try {
+        return cb();
+    } finally {
+        clock.restore();
+    }
+}
+
+describe("SessionClaimValidator Refresh", () => {
```
porcellus, line 33: These are fine for testing the interaction between the `DateProvider` and claim validation, but we also need to test the clock skew detection mechanisms. Please add some tests "cross-interception" tests. See `test/interception.claims.test.js`

test/sessionClaimValidatorRefresh.test.js
```diff
@@ -0,0 +1,129 @@
...
+                refresh: () => {
+                    tokenPayload["st-test"].t = Date.now();
+                },
+                defaultMaxAgeInSeconds: 10 /* 10 seconds */
+            });
+
+            const testClaimValidator = testClaim.validators.isTrue();
+
+            // Adjust DateProvider clientClockDeviation
+            DateProviderReference.getReferenceOrThrow().dateProvider.setClientClockDeviationInMillis(-ONE_HOUR_IN_MS);
```
porcellus, line 116: You can move this into the `refresh` function to "simulate" this being done by response interception.

test/sessionClaimValidatorRefresh.test.js
```diff
@@ -0,0 +1,129 @@
...
+            assert(testClaimValidator.shouldRefresh(tokenPayload) === false);
+        });
+    });
+
+    describe("Client Clock is ahead 1 hour of server time", () => {
+        before(function () {
+            DateProviderReference.init();
+        });
+
+        it("SessionClaimValidator.shouldRefresh should return true even after refreshing", function () {
```
porcellus, line 68: add something like " without updating date provider" into the case name. (and the opposite to the other test)

porcellus (changes requested):

lib/ts/utils/dateProvider/index.ts
anku255: Thanks for catching that. I have fixed it.

My manual tests passed because it was getting initialized by supertokens-web-js.

test/sessionClaimValidatorRefresh.test.js
anku255: done

lib/ts/fetch.ts
anku255: done

lib/ts/recipeImplementation.ts
anku255: done

lib/ts/utils/dateProvider/defaultImplementation.ts
anku255: I did this but it means we have to read from localStorage everytime we call `getClientClockSkewInMillis`. I have explained the reason in a comment.

test/sessionClaimValidatorRefresh.test.js
anku255: done

test/sessionClaimValidatorRefresh.test.js
anku255: done

lib/ts/axios.ts
```diff
@@ -188,6 +189,18 @@ export function responseInterceptor(axiosInstance: any) {
             const preRequestLSS = await getLocalSessionState(false);
             await saveTokensFromHeaders(response);
 
+            const frontToken = response.headers["front-token"];
```
porcellus, line 192: Can this be undefined? If so, it may be safer to also expect & check for that in `updateClockSkewUsingFrontToken`

lib/ts/fetch.ts
```diff
@@ -926,3 +927,38 @@ export function fireSessionUpdateEventsIfNecessary(
         });
     }
 }
+
+/**
+ * Updates the clock skew based on the provided frontToken and theround-trip time.
```
porcellus, line 932: This should be updated

lib/ts/fetch.ts
```diff
@@ -926,3 +927,38 @@ export function fireSessionUpdateEventsIfNecessary(
...
+ *
+ * @param {Object} params - The parameters for updating the clock skew.
+ * @param {string | null} params.frontToken - The frontToken containing issued timestamp.
+ * @param {string | null} params.responseHeaders - The headers for the response that returned the frontToken.
```
porcellus, line 936: The type here is wrong. Isn't there a way to omit it? I like the comment, but we already have the type because of TS.

lib/ts/fetch.ts
```diff
@@ -926,3 +927,38 @@ export function fireSessionUpdateEventsIfNecessary(
...
+    responseHeaders
+}: {
+    frontToken: string | null;
+    responseHeaders: Headers;
+}): void => {
+    logDebugMessage("updateClockSkewUsingFrontToken: frontToken: " + frontToken);
+
+    if (frontToken === null || frontToken === "remove") {
+        logDebugMessage(
+            "updateClockSkewUsingFrontToken: frontToken is either null or is being removed, skipping update"
```
porcellus, line 949:
```suggestion
            "updateClockSkewUsingFrontToken: the access token payload wasn't updated or is being removed, skipping clock skew update"
```

lib/ts/fetch.ts
```diff
@@ -926,3 +927,38 @@ export function fireSessionUpdateEventsIfNecessary(
...
+        return;
+    }
+
+    const frontTokenPayload = parseFrontToken(frontToken);
+    const clockSkewInMillis = AuthHttpRequest.recipeImpl.getClockSkewInMillis({
+        accessTokenPayload: frontTokenPayload.up,
+        responseHeaders
+    });
+
+    console.log("clockSkewInMillis", clockSkewInMillis);
```
porcellus, line 960:
```suggestion

```

lib/ts/fetch.ts
```diff
@@ -926,3 +927,38 @@ export function fireSessionUpdateEventsIfNecessary(
...
+ *
+ * @param {Object} params - The parameters for updating the clock skew.
+ * @param {string | null} params.frontToken - The frontToken containing issued timestamp.
+ * @param {string | null} params.responseHeaders - The headers for the response that returned the frontToken.
+ */
+export const updateClockSkewUsingFrontToken = ({
+    frontToken,
+    responseHeaders
+}: {
+    frontToken: string | null;
```
porcellus, line 942: I think it's better to allow undefined here.

lib/ts/version.ts
```diff
@@ -12,6 +12,6 @@
  * License for the specific language governing permissions and limitations
  * under the License.
  */
-export const package_version = "17.0.5";
+export const package_version = "17.0.6";
```
porcellus, line 15: This should be a minor version increase (it's a new feature, but it's backwards compatible, not a fix).

lib/ts/utils/dateProvider/defaultImplementation.ts
```diff
@@ -0,0 +1,38 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+import WindowHandlerReference from "../windowHandler";
+
+// To avoid potential issues with initializing clockSkewInMillis as a class variable in the constructor due to WindowHandlerReference not being initialized at DateProvider instance creation,
+// we read from localStorage each time `getClientClockSkewInMillis` is called.
```
porcellus, line 19: Why can't we do this in `DateProviderReference` constructor? That should be safe.

lib/ts/recipeImplementation.ts
```diff
@@ -308,6 +308,27 @@ export default function RecipeImplementation(recipeImplInput: {
...
+            accessTokenPayload: any;
+            responseHeaders: Headers;
+        }): number {
+            logDebugMessage("getClockSkewInMillis: called");
+
+            const tokenIssuedAt = accessTokenPayload?.iat;
+            if (tokenIssuedAt === undefined || typeof tokenIssuedAt !== "number") {
+                logDebugMessage("getClockSkewInMillis: payload iat is undefined or not a number, returning 0");
+                return 0;
+            }
```
porcellus, line 325: This should never happen, so we can throw an error here. (That way test failures will let us know if this is something we do need to worry about)
This information should be present in all access tokens I think. (maybe under a different name for older version)

lib/ts/recipeImplementation.ts
```diff
@@ -308,6 +308,27 @@ export default function RecipeImplementation(recipeImplInput: {
...
+
+        getClockSkewInMillis: function ({
+            accessTokenPayload
+        }: {
+            accessTokenPayload: any;
+            responseHeaders: Headers;
+        }): number {
+            logDebugMessage("getClockSkewInMillis: called");
+
+            const tokenIssuedAt = accessTokenPayload?.iat;
```
porcellus, line 321: Check if this works with st-node version 14. I think that is still supported by this SDK and the access token payload format changed in version 15. We may need to add a fallback here to make this work with old versions.

test/interception.claims.test.js
```diff
@@ -191,5 +191,234 @@ addGenericTestCases((name, transferMethod, setupFunc, setupArgs = []) => {
                 await browser.close();
             }
         });
+
+        it("should call the claim refresh endpoint once for multiple `shouldRefresh` calls with adjusted clock skew", async function () {
```
porcellus, line 195: just double checking: this fails with older versions, right?

test/interception.claims.test.js
```diff
@@ -191,5 +191,234 @@ addGenericTestCases((name, transferMethod, setupFunc, setupArgs = []) => {
...
+                assert.strictEqual(customClaimRefreshCalledCount, 1);
+            } finally {
+                await browser.close();
+            }
+        });
+    });
+});
+
+addGenericTestCases((name, transferMethod, setupFunc, setupArgs = []) => {
+    describe(`${name}: SessionClaimValidator Refresh With Clock Skew`, function () {
```
porcellus, line 272: This test case is just checking if you disable the clock skew the other test would break, right? If so we do not need this. I'm fine with you confirming that the above testcase fails if you run it with an older version of the SDK.

test/interception.testgen.js
```diff
@@ -13,6 +13,12 @@ module.exports.addGenericTestCases = function (getTestCases) {
                                 shouldDoInterceptionBasedOnUrl: url => url.includes("doOverride")
                             }));
                         }
+                        if (reqOverride === "getClockSkewInMillis") {
+                            overrideFunctions.push(oI => ({
+                                ...oI,
+                                getClockSkewInMillis: () => 0
+                            }));
+                        }
```
porcellus, line 21: I don't think we'll need this.

porcellus (changes requested):

test/server/index.js
```diff
@@ -18,6 +18,7 @@ let SessionRecipeRaw = require("supertokens-node/lib/build/recipe/session/recipe
 let Querier = require("supertokens-node/lib/build/querier").Querier;
 let NormalisedURLPath = require("supertokens-node/lib/build/normalisedURLPath").default;
 let Session = require("supertokens-node/recipe/session");
+let { BooleanClaim } = require("supertokens-node/recipe/session/claims");
```
anku255, line 21:
```suggestion
```

lib/ts/axios.ts
porcellus, line 192: The front token is saved in multiple places, but I only see a single call to update the clock skew. It may be better if you moved this into `saveTokensFromHeaders` (for all interceptors, or into a place more centralized). We want to make sure we are not making a mistake by updating the front token but skipping the clock skew update.

lib/ts/recipeImplementation.ts
```diff
@@ -308,6 +308,26 @@ export default function RecipeImplementation(recipeImplInput: {
...
+            if (tokenIssuedAt === undefined || typeof tokenIssuedAt !== "number") {
+                throw new Error("payload iat is undefined or not a number");
```
porcellus, line 323: As discussed in the sprint meeting, please update this to only log the fact that the clock skew is not updated because of an unsupported BE sdk version

lib/ts/utils/dateProvider/index.ts
```diff
@@ -0,0 +1,63 @@
...
+        let dateProviderFunc: DateProviderInput = original => original;
+        if (dateProviderInput !== undefined) {
+            dateProviderFunc = dateProviderInput;
+            this.dateProvider = dateProviderFunc(defaultDateProviderImplementation);
+        } else {
+            this.dateProvider = dateProviderFunc(defaultDateProviderImplementation);
+            // Initialise the clockSkewInMillis from localStorage for the default implementation.
+            const localStorage = WindowHandlerReference.getReferenceOrThrow().windowHandler.localStorage;
+            const clockSkewInMillis = parseInt(localStorage.getItemSync(DateProvider.CLOCK_SKEW_KEY) || "0", 10);
+            this.dateProvider.setClientClockSkewInMillis(clockSkewInMillis);
```
porcellus, line 42: I'd prefer if this was done in a factory function (a static `init`) in the default date provider impl class.
We can then call that factory function here instead of exporting an instance from there.
Please also comment that we use an `init` call to build the instance because we need it to execute after the window handler is initialized.

CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [unreleased]
 
+## [17.0.6] - 2024-01-11
```
porcellus, line 10: The version number was not updated here.

porcellus (changes requested):
Please ensure (pre-release checklist):
- [ ] All tests are passing
- [ ] New tests fail when running with the old st-website version
- [ ] Check that the original issue (the email verification validation loop) is solved by trying to replicate it the same way you originally did (changing your system clock)

Please manually test that the above email verification issue is gone and the auth-react test app is generally usable while you have misaligned system clock, plus: 
- [ ] slow connection (i.e.: update the connection speed in the browser debugger)
- [ ] fast connection (which is normal when using localhost)
- [ ] slow server response (add different delays before and after the payload update)
- [ ] system clock update while the page is open

lib/ts/axios.ts
anku255: Good observation! I have moved it to `saveTokensFromHeaders` for all the interceptors.

lib/ts/recipeImplementation.ts
anku255: done

lib/ts/utils/dateProvider/index.ts
anku255: I have made the changes but please do check if that is what you were trying to suggest.

CHANGELOG.md
anku255: done

porcellus (approved):

lib/ts/utils/dateProvider/defaultImplementation.ts
```diff
@@ -0,0 +1,59 @@
...
+        if (DateProvider.instance === undefined) {
+            throw new Error("DateProvider must be initialized before calling this method.");
+        }
+
+        return DateProvider.instance;
+    }
+
+    setClientClockSkewInMillis(clockSkewInMillis: number): void {
+        this.clockSkewInMillis = clockSkewInMillis;
+        const localStorage = WindowHandlerReference.getReferenceOrThrow().windowHandler.localStorage;
```
rishabhpoddar, line 48: i don't think we should use localstorage cause of people sharing session across sub domains. So we should:
- use frontend cookies; OR
- ideally, no need to store this value in a storage, and just store in memory. If it's not there in memory, can't we calculate it based on the access token payload and then save it in memory?

rishabhpoddar (changes requested):

lib/ts/utils/dateProvider/defaultImplementation.ts
anku255: Upon several discussions, we decided to keep the caching strategy while documenting all the considered cases.

rishabhpoddar (approved):

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "supertokens-website",
-  "version": "17.0.5",
+  "version": "18.0.0",
   "description": "frontend sdk for website to be used for auth solution.",
   "main": "index.js",
```
rishabhpoddar, line 5: size pr check is failing

rishabhpoddar (changes requested):

PR #242, opened 2024-01-20
feat: add threshold in DateProvider

anku255:
## Summary of change

This PR adds a threshold in DateProvider to fix some breaking tests and ensure we only have clock skew when its significant enough.

## Test Plan

## Documentation changes

   - In `package.json`
   - In `package-lock.json`
   - In `lib/ts/version.ts`


lib/ts/claims/primitiveArrayClaim.ts
```diff
@@ -33,6 +33,11 @@ export class PrimitiveArrayClaim<ValueType> {
             id?: string
         ): SessionClaimValidator => {
             const DateProvider = DateProviderReference.getReferenceOrThrow().dateProvider;
+            if (maxAgeInSeconds && maxAgeInSeconds < DateProvider.getThresholdInSeconds()) {
```
rishabhpoddar, line 36: no implicit checks! change it to maxAgeInSeconds !== undefined. What if the user puts maxAgeInSeconds to 0? Then if (maxAgeInSeconds) will be false.

lib/ts/utils/dateProvider/types.ts
```diff
@@ -14,9 +14,10 @@
  */
 
 export interface DateProviderInterface {
+    getThresholdInSeconds(): number;
     now(): number;
     setClientClockSkewInMillis(clockSkewInMillis: number): void;
     getClientClockSkewInMillis(): number;
 }
```
rishabhpoddar, line 21: we also need to allow the user to set their own threshold value. So make this a config please.

rishabhpoddar (changes requested):

test/interception.claims.test.js
```diff
@@ -266,6 +266,78 @@ addGenericTestCases((name, transferMethod, setupFunc, setupArgs = []) => {
             }
         });
 
+        it("should call the claim refresh endpoint once for multiple `shouldRefresh` calls with adjusted clock skew (client clock behind)", async function () {
```
rishabhpoddar, line 269: you should add to this test that if there is no clock skew correction, the shouldRefetch always returns false, even when it should return true. And then after the clock skew correction, it behaves correctly.

rishabhpoddar (changes requested):

lib/ts/utils/dateProvider/types.ts
```diff
@@ -14,9 +14,10 @@
  */
 
 export interface DateProviderInterface {
+    getThresholdInSeconds(): number;
     now(): number;
     setClientClockSkewInMillis(clockSkewInMillis: number): void;
     getClientClockSkewInMillis(): number;
```
rishabhpoddar, line 20: there is also a recipe function called `getClockSkewInMillis`, which is kind of confusing given that these two functions also exist. From a user's point of view, which one should they call? DateProvider. getClientClockSkewInMillis(), or sessionrecipe.getClockSkewInMillis() (from a naming point of view, it's confusing).

So, i think you should come up with a better name for the recipe interface function: `getClockSkewInMillis`  or restructure how stuff is done..

rishabhpoddar (changes requested):

lib/ts/claims/primitiveArrayClaim.ts
anku255: Thanks for catching it. I have fixed it. There must an eslint rule to prevent this.. but I we don't seem to be using eslint in this project currently.

lib/ts/utils/dateProvider/types.ts
anku255: Yeah we discussed this and decided against it as we thought, one can simply override the DateProvider if they want to do this.

But I have added it back as the overriding function doesn't get originaImplementation anymore.

test/interception.claims.test.js
anku255: Well, I had this test added intially - [see this ](https://github.com/supertokens/supertokens-website/pull/241/commits/5a69ae7917578f7f517c74779cf1a94b416ab104#diff-281370ee13572ac2d979d348000821f73c381622751fbbf1d8fd40554052704bR350-R422) but @porcellus felt it was [not needed ](https://github.com/supertokens/supertokens-website/pull/241/files/5a69ae7917578f7f517c74779cf1a94b416ab104#r1449537639) for the amount of changes it requires in `interception.testgen.ts` and I removed it afterwards.

I have no strong feeling and can add it back if you prefer it that way.

lib/ts/utils/dateProvider/types.ts
anku255: Renamed it to `calculateClockSkewInMillis`. I think that's a more appropriate name.

porcellus: Correct me if I'm wrong, but `getClockSkewInMillis` is not exposed to users from the recipe (partially because of this)

anku255: It is not exposed but users can override it. So, I renamed it to avoid any confusion.

test/interception.claims.test.js
porcellus: The issue isn't the amount of required changes.
My reasoning was that basically overriding to disable a fix (which is done in the `beforeEach` part of the suite) and then testing that things go wrong isn't required. We usually just create a new test that fails with the previous version (which this one does)

porcellus: After a quick discussion with @rishabhpoddar:
- we do not need to re-add the test that disables the fix
- please confirm (e.g.: as an answer to this comment) that the new test fail with the old version
- describe the old behaviour (and how/why it fails the new tests)

lib/ts/utils/dateProvider/types.ts
porcellus: Ok, just making sure.

test/interception.claims.test.js
anku255: I replicated the test file into the master branch and executed the tests.

- Tests fail when the client clock is 1 hour ahead.
- Tests pass when the client clock is 1 hour behind, as expected.

Within the `validateClaims` function, each validator's `shouldRefresh` function is called prior to claim validation. If the `shouldRefresh` function returns `true`, the claim value is refreshed through the designated function. The `shouldRefresh` function assesses the time elapsed since the token's last update by the backend. If this duration exceeds the specified threshold, `true` is returned. If no existing claim data is present, `shouldRefresh` defaults to `true`.

When the client clock is ahead of the server clock by more than the threshold, `shouldRefresh` consistently returns `true`, causing the test failures in the master branch. Conversely, if the client clock lags behind, `shouldRefresh` returns `false`.

In the current branch, the `shouldRefresh` function adopts `DateProvider.now` for comparisons instead of `Date.now`. `DateProvider` accommodates the discrepancy between client and server clocks, ensuring accurate comparisons and resolving the test failures.

porcellus: Ideally the "behind" test should fail as well in the older version. Please try and change it so. One way to do it is to intentionally set "outdated" claim values in the refresh (instead of setting a shorter timeout and waiting) and checking if the refresh function was called all 3 times. This should be true in the new version but not in the old one.

anku255: done

porcellus (approved):

lib/ts/claims/primitiveArrayClaim.ts
```diff
@@ -131,7 +131,7 @@ export class PrimitiveArrayClaim<ValueType> {
             id?: string
         ): SessionClaimValidator => {
             const DateProvider = DateProviderReference.getReferenceOrThrow().dateProvider;
-            if (maxAgeInSeconds && maxAgeInSeconds < DateProvider.getThresholdInSeconds()) {
+            if (maxAgeInSeconds !== undefined && maxAgeInSeconds < DateProvider.getThresholdInSeconds()) {
```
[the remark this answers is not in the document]
anku255: done

rishabhpoddar (approved):

PR #243, opened 2024-01-22
chore: add test for checking interception for relative URL

### LINKED ISSUE #238: Support interception if request made to relative path
Right now, if there is no domain in the request, and is being made to a relative path, it will not do the network interception even though the apiDomain configuration might be equal to the loaded URL on the browser. So we should instead check the current URL on the browser as opposed to not doing interception at all.

prateek3255:
## Summary of change
Add test for checking interception for relative URL

## Related issues
Fixes #238 

## Test Plan
Have verified with a local demo as well

## Documentation changes
N/A

   - In `package.json`
   - In `package-lock.json`
   - In `lib/ts/version.ts`

N/A


test/fetch.test.js
```diff
@@ -2325,4 +2325,40 @@ describe("Fetch AuthHttpRequest class tests", function () {
...
+                    headers: {
+                        Accept: "application/json",
+                        "Content-Type": "application/json"
+                    },
+                    body: JSON.stringify({ userId })
+                });
+
+                assertEqual(await loginResponse.text(), userId);
+
+                let getResponse = await fetch(`/check-rid`);
```
rishabhpoddar, line 2356: add similar test with axios and xhr as well please.

rishabhpoddar (changes requested):

test/fetch.test.js
prateek3255: Done

rishabhpoddar (approved):

PR #258, opened 2024-05-31
fix: session refresh loop in all request interceptors

anku255:
## Summary of change

This PR fixes the session refresh loop in all the request interceptors that occurred when an API returned a 401 response despite a valid session. Interceptors now attempt to refresh the session a maximum of three times before returning the last API response. The retry limit is configurable via the `maxRetryAttemptsForSessionRefresh` option.

## Test Plan

## Documentation changes

   - In `package.json`
   - In `package-lock.json`
   - In `lib/ts/version.ts`


lib/ts/axios.ts
```diff
@@ -463,7 +463,26 @@ export default class AuthHttpRequest {
...
+                            /**
+                             * An API may return a 401 error response even with a valid session, causing a session refresh loop in the interceptor.
+                             * To prevent this infinite loop, we break out of the loop after retrying the original request a specified number of times.
+                             * The maximum number of retry attempts is defined by maxRetryAttemptsForSessionRefresh config variable.
+                             */
+                            if (hasExceededMaxSessionRefreshAttempts(config)) {
+                                logDebugMessage(
+                                    `doRequest: Maximum session refresh attempts reached. sessionRefreshAttempts: ${config.__supertokensSessionRefreshAttempts}, maxRetryAttemptsForSessionRefresh: ${AuthHttpRequestFetch.config.maxRetryAttemptsForSessionRefresh}`
+                                );
+                                return response;
```
porcellus, line 476: After some discussion I think our interceptors should throw a JS error explaining that this is a refresh loop and any information you think is relevant.

This way it'll be more visible to devs and there should be no way of ignoring this in development.
This means that the site will still break in production if there is a refresh loop, but:
- it used to break anyway
- it doesn't increase the server load as much
- it is much more visible to devs (through error logs) and to users who can report this issue to the devs

Please update the interceptors/tests/changelogs.

lib/ts/utils/index.ts
```diff
@@ -116,6 +116,14 @@ export function validateAndNormaliseInputOrThrowError(options: InputType): Norma
         sessionTokenBackendDomain = normaliseSessionScopeOrThrowError(options.sessionTokenBackendDomain);
     }
 
+    let maxRetryAttemptsForSessionRefresh = 3;
```
porcellus, line 119: Let's make this 10 to be safe.

porcellus (changes requested):

lib/ts/utils/index.ts
anku255: Done

lib/ts/axios.ts
anku255: Done

lib/ts/axios.ts
```diff
@@ -430,27 +450,7 @@ export default class AuthHttpRequest {
...
-                        }
-                        logDebugMessage("doRequest: Retrying original request");
-                    } else {
-                        if (response.status === AuthHttpRequestFetch.config.invalidClaimStatusCode) {
-                            await onInvalidClaimResponse(response);
-                        }
-
-                        return response;
-                    }
+                    return response;
```
porcellus, line 453: Please add a comment here explaining why we do not need to check the status here.

lib/ts/xmlhttprequest.ts
```diff
@@ -240,7 +248,12 @@ export function addInterceptorsToXMLHttpRequest() {
                 } else {
                     // Here we only need to handle fetch related errors, from the refresh endpoint called by the retry
                     // So we should only get network level errors here
-                    redispatchEvent("error", new Event("error"));
+
+                    const ev = new Event("error");
+                    (ev as any).error = err;
+                    self.dispatchEvent(ev);
+
+                    redispatchEvent("error", ev);
```
porcellus, line 256: This looks strange, why do we need both calls?

porcellus (changes requested):

lib/ts/axios.ts
anku255: done.

lib/ts/xmlhttprequest.ts
anku255: We decided to call the `onerror` method directly, instead.

porcellus (approved):

lib/ts/axios.ts
```diff
@@ -463,7 +467,29 @@ export default class AuthHttpRequest {
...
+                                );
+
+                                const errorMessage = `Received a 401 response from ${url}. Attempted to refresh the session and retry the request with the updated session tokens ${AuthHttpRequestFetch.config.maxRetryAttemptsForSessionRefresh} times, but each attempt resulted in a 401 error. The maximum session refresh limit has been reached. Please investigate your API. To increase the session refresh attempts, update maxRetryAttemptsForSessionRefresh in the config.`;
+                                console.error(errorMessage);
+                                throw new Error(errorMessage);
+                            }
+
                             const refreshResult = await onUnauthorisedResponse(preRequestLSS);
+
+                            incrementSessionRefreshAttemptCount(config);
```
rishabhpoddar, line 488: why do we increment this even if refresh failed?

rishabhpoddar (changes requested):

rishabhpoddar (approved):

PR #262, opened 2024-06-26
fix: session refresh loop if cookies writes are disabled

anku255:
## Summary of change

This PR fixes the session refresh loop if cookies writes are disabled.

## Test Plan

## Documentation changes

   - In `package.json`
   - In `package-lock.json`
   - In `lib/ts/version.ts`


porcellus (approved):

lib/ts/fetch.ts
```diff
@@ -668,8 +668,24 @@ export async function getLocalSessionState(tryRefresh: boolean): Promise<LocalSe
                     status: "NOT_EXISTS"
                 };
             }
-            logDebugMessage("getLocalSessionState: Retrying post refresh");
-            return await getLocalSessionState(tryRefresh);
+
+            const lastAccessTokenUpdate = await getFromCookies(LAST_ACCESS_TOKEN_UPDATE);
```
rishabhpoddar, line 672: this could come here if there is an API_ERROR output from refresh as well, but thats not handled here?

lib/ts/axios.ts
```diff
@@ -597,7 +599,10 @@ async function saveTokensFromHeaders(response: AxiosResponse) {
 
     const antiCsrfToken = response.headers["anti-csrf"];
     if (antiCsrfToken !== undefined) {
-        const tok = await getLocalSessionState(true);
+        // Call getLocalSessionState with tryRefresh: false as the session was just refreshed.
+        // If the local session doesn't exist, it means we failed to write to cookies.
+        // Using tryRefresh: true would cause an infinite refresh loop.
```
rishabhpoddar, line 604: i dont think this comment is right. How is it that if it comes here, it means the session was just refreshed? It could come here even if a new session is just created.

rishabhpoddar: And therefore, why is the change to pass false instead of true correct?

lib/ts/axios.ts
```diff
@@ -243,7 +243,9 @@ export function responseInterceptor(axiosInstance: any) {
                 !doNotDoInterception &&
                 // we do not call doesSessionExist here cause the user might override that
                 // function here and then it may break the logic of our original implementation.
-                !((await getLocalSessionState(true)).status === "EXISTS")
+
+                // Calling getLocalSessionState with tryRefresh: false, since the session would have been refreshed in the try block if expired.
+                (await getLocalSessionState(false)).status === "NOT_EXISTS"
             ) {
```
rishabhpoddar, line 249: should we get rid of this if block entirely? Cause if the session doesn't exist, it means the front token and anti csrf token should already be removed?

lib/ts/fetch.ts
```diff
@@ -797,7 +813,10 @@ async function saveTokensFromHeaders(response: Response) {
     }
     const antiCsrfToken = response.headers.get("anti-csrf");
     if (antiCsrfToken !== null) {
-        const tok = await getLocalSessionState(true);
+        // Call getLocalSessionState with tryRefresh: false as the session was just refreshed.
+        // If the local session doesn't exist, it means we failed to write to cookies.
+        // Using tryRefresh: true would cause an infinite refresh loop.
+        const tok = await getLocalSessionState(false);
```
rishabhpoddar, line 819: see comment in axios.ts file related to this change

lib/ts/xmlhttprequest.ts
```diff
@@ -70,6 +66,21 @@ export function addInterceptorsToXMLHttpRequest() {
 
         const eventHandlers: Map<keyof XMLHttpRequestEventMap, Set<XHREventListener<any>>> = new Map();
 
+        let delayedQueue = firstEventLoopDone;
```
rishabhpoddar, line 69: can you explain this? Add comments here in code too.

lib/ts/xmlhttprequest.ts
```diff
@@ -70,6 +66,21 @@ export function addInterceptorsToXMLHttpRequest() {
 
         const eventHandlers: Map<keyof XMLHttpRequestEventMap, Set<XHREventListener<any>>> = new Map();
```
rishabhpoddar, line 68: we have changed the implementation of saveTokensFromHeaders in axios.ts and fetch.ts. Why not in this file?

rishabhpoddar (changes requested):

lib/ts/fetch.ts
porcellus: It only gets here if the result is `RETRY` (see ln 664)

lib/ts/axios.ts
porcellus: We do not know for sure if the anti-csrf one is removed or not, but the front-token shouldn't exist here anyway.

lib/ts/fetch.ts
```diff
@@ -668,8 +668,24 @@ export async function getLocalSessionState(tryRefresh: boolean): Promise<LocalSe
...
+
+            const lastAccessTokenUpdate = await getFromCookies(LAST_ACCESS_TOKEN_UPDATE);
+            const frontTokenExists = await FrontToken.doesTokenExists();
+
+            // If we fail to retrieve the local session state after a successful refresh,
+            // it indicates an issue with writing to cookies. Without the FrontToken,
+            // session refresh may work but other SDK functionalities won't work as expected.
+            // Therefore, we throw an error here instead of retrying.
+            if (!frontTokenExists || lastAccessTokenUpdate === undefined) {
+                const errorMessage = `Failed to retrieve local session state from cookies after a successful session refresh. This indicates a configuration error or that the browser is preventing cookie writes (e.g., incognito mode).`;
```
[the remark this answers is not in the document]
anku255: Updated the error messages to remove the incognito part as that is not true for most browsers.  I couldn't find a general example other than a user explicitly  disabling it in browser setting. I believe our message is generic enough to accommodate that.

lib/ts/axios.ts
anku255: I have updated the comment to include the newly created session case.. However, the point is that we don't want to call `getLocalSessionState` with tryRefresh: true there because it may cause a refresh loop and since its not needed, we simply call it with false.

lib/ts/fetch.ts
anku255: Updated it as well.

lib/ts/xmlhttprequest.ts
anku255: Added comments but basically we needed to do that to make XHR throw. Previously it was just logging the error to console.

lib/ts/xmlhttprequest.ts
anku255: Sorry, that was a miss. I have fixed it now.

lib/ts/xmlhttprequest.ts
rishabhpoddar: I dont see any comments.. not sure what you added. I have no idea what this change in XHR is supposed to even mean.

anku255: The comment in on line 73, because that's what changed.

Line 69-70 has been moved down. They haven't changed.

anku255: Previously, we were calling `console.error` if any error occurred in `delayIfNecessary` function but that would supress the error, instead we wanted it to throw. That's why I added the call to `onerror`.

rishabhpoddar (approved):

PR #263, opened 2024-07-03
fix: failiing tests

anku255:
## Summary of change

This PR has a couple of changes - 

### 1. "test sameSite is none if using iframe" test fix

In #262, we changed `getLocalSessionState` to be called with `tryRefresh: false` wherever it wasn't required (such as just after a session was created/refreshed). 

This removed unnecessary calls to the refresh API in certain cases. One such is when the FrontToken doesn't get saved (either because of SameSite: None; Secure setting on HTTP non-localhost domain or because of cookie writes being disabled).

The "test sameSite is none if using iframe"  test relied on this refresh call. After its removal the test started failing.

The test implementation wasn't correct. It was relying on a side-effect. When `isInIframe` is set to true, we set the Frontend cookies with `SameSite: true; Secure`. Since the test domain was `http://localhost.org:8080`, the browsers simply ignored all the frontend cookies such as anticsrf and frontToken cookie.  Prior to #262, we would do a refresh call just after login, this removed all the backend cookies as well (because antiCsrf token was not present). The test was checking that all the cookies would be removed. After #262, the backend cookies were not removed because the /refresh call never happened.

The correct test implementation should have been checking the `sameSite` and `secure` attributes of the frontend cookies. However, for this we needed to make sure the domain was localhost and not localhost.org.  This PR fixes this and also updates all the tests.

### 2. Refresh loop while calling the /refresh API is called when front token is not available

While debugging the tests, we found out that if session is not available and a /refresh API is called directly then it would cause a refresh loop (upto 10 times by default).

If there is no session, then `preRequestLSS.status` would be `NOT_EXISTS` but `postLockLSS.status` would be `MAY_EXISTS`.  In `onUnauthorisedResponse`, the condition `preRequestLSS.status !== postLockLSS.status` was always true, so it kept looping. This has been fixed.

## Related issues
- Link to issue1 here
- Link to issue1 here

## Test Plan

## Documentation changes

   - In `package.json`
   - In `package-lock.json`
   - In `lib/ts/version.ts`


test/interception.basic1.test.js
```diff
@@ -227,8 +238,23 @@ addTestCases((name, transferMethod, setupFunc, setupArgs = []) => {
...
+            // The following cookies may or may not be set depending on header or cookie based auth.
+            // We will check the presence of the cookies before asserting the attributes
+            const frontendCookies = ["sAntiCsrf", "st-last-access-token-update", "st-access-token", "st-refresh-token"];
+            frontendCookies.forEach(cookieName => {
+                const cookie = cookies.find(cookie => cookie.name === cookieName);
+                if (cookie) {
+                    assert(cookie.sameSite === "None");
+                    assert(cookie.secure === true);
+                }
+            });
```
porcellus, line 257: Please make these into an explicit check, we know which we expect to be there and which we don't based on the  transfermethod

test/fetch.test.js
```diff
@@ -501,10 +501,12 @@ describe("Fetch AuthHttpRequest class tests", function () {
         });
         try {
             const page = await browser.newPage();
-            await page.goto(BASE_URL + "/index.html", { waitUntil: "load" });
+            // NOTE: Using localhost:8080 as the base URL because browsers ignore
```
porcellus, line 504: Please confirm that these tests are duplicated by the test generator and remove them from the "legacy" test files.

porcellus (changes requested):

test/interception.basic1.test.js
anku255: Didn't realize we had access to `transferMethod` in the test. I have updated the test to make explicit checks

test/fetch.test.js
anku255: Yup, these were indeed duplicated. I have removed them.

porcellus (approved):

